PR #6, opened 2017-04-10
Resolve middleware server status from availability metric

israel-hdez:
Server status was being fetched from inventory's data resulting in wrong
reported status when server is down. Now, server status is resolved with
a combination of server's availability metric and its inventory status.

https://bugzilla.redhat.com/show_bug.cgi?id=1438816

This PR goes together with manageiq/manageiq-ui-classic#987. Merge order is not important and is not required to merge simultaneously.

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_deployment.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_deployment.rb
@@ -1,4 +1,14 @@
 module ManageIQ::Providers
   class Hawkular::MiddlewareManager::MiddlewareDeployment < MiddlewareDeployment
+    def self.resource_path_for_metrics(item)
+      path = ::Hawkular::Inventory::CanonicalPath.parse(item.ems_ref)
+      # for subdeployments use it's parent deployment availability.
+      path = path.up if path.resource_ids.last.include? CGI.escape('/subdeployment=')
+      # Ensure consistency on keys (resource_path) used on metric_id_by_resource_path
+      path = ::Hawkular::Inventory::CanonicalPath.new(:tenant_id    => path.tenant_id,
+                                                      :feed_id      => path.feed_id,
+                                                      :resource_ids => path.resource_ids)
+      path.to_s
+    end
   end
 end
--- a/spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
+++ b/spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
@@ -26,6 +26,7 @@ describe ManageIQ::Providers::Hawkular::Inventory::Parser::MiddlewareManager do
     parser.persister = persister_double
     parser
   end
+  let(:stubbed_metric_data) { OpenStruct.new(:id => 'm1', :data => [{'timestamp' => 1, 'value' => 'arbitrary value'}]) }
   let(:server) do
     FactoryGirl.create(:hawkular_middleware_server,
                        :name                  => 'Local',
@@ -33,7 +34,8 @@ describe ManageIQ::Providers::Hawkular::Inventory::Parser::MiddlewareManager do
                        :ems_ref               => '/t;Hawkular'\
                                                  "/f;#{the_feed_id}/r;Local~~",
                        :nativeid              => 'Local~~',
-                       :ext_management_system => ems_hawkular)
+                       :ext_management_system => ems_hawkular,
+                       :properties            => { 'Server Status' => 'Inventory Status' })
   end
 
   describe 'parse_datasource' do
@@ -107,7 +109,6 @@ describe ManageIQ::Providers::Hawkular::Inventory::Parser::MiddlewareManager do
   describe 'fetch_availabilities_for' do
     let(:stubbed_resource) { OpenStruct.new(:manager_uuid => '/t;hawkular/f;f1/r;stubbed_resource') }
     let(:stubbed_metric_definition) { OpenStruct.new(:path => '/t;hawkular/f;f1/r;stubbed_resource/m;m1', :hawkular_metric_id => 'm1') }
-    let(:stubbed_metric_data) { OpenStruct.new(:id => 'm1', :data => [{'timestamp' => 1, 'value' => 'arbitrary value'}]) }
 
     before do
       allow(collector_double).to receive(:metrics_for_metric_type).and_return([])
@@ -178,7 +179,6 @@ describe ManageIQ::Providers::Hawkular::Inventory::Parser::MiddlewareManager do
 
   describe 'fetch_deployment_availabilities' do
     let(:stubbed_deployment) { OpenStruct.new(:manager_uuid => '/t;hawkular/f;f1/r;s1/r;d1') }
-    let(:stubbed_metric_data) { OpenStruct.new(:id => 'm1', :data => [{'timestamp' => 1, 'value' => 'up'}]) }
 
     before do
       allow(persister_double).to receive(:middleware_deployments).and_return([stubbed_deployment])
@@ -193,6 +193,8 @@ describe ManageIQ::Providers::Hawkular::Inventory::Parser::MiddlewareManager do
     end
 
     it 'assigns enabled status to a deployment with "up" metric' do
+      stubbed_metric_data.data.first['value'] = 'up'
+
       parser.fetch_deployment_availabilities(%w(f1))
       expect(stubbed_deployment.status).to eq('Enabled')
     end
@@ -205,8 +207,6 @@ describe ManageIQ::Providers::Hawkular::Inventory::Parser::MiddlewareManager do
     end
 
     it 'assigns unknown status to a deployment whose metric is something else than "up" or "down"' do
-      stubbed_metric_data.data.first['value'] = 'this is arbitrary value'
-
       parser.fetch_deployment_availabilities(%w(f1))
       expect(stubbed_deployment.status).to eq('Unknown')
     end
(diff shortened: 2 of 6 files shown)


israel-hdez:
@rubenvp8510 @cfcosta may you review?

israel-hdez:
@miq-bot assign @abonas

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
```diff
@@ -97,79 +97,149 @@ def inventory_object_data(inventory_object)
...
+      # It will return five metrics for deployments. Relationships metric_id => deployment_ids are:
+      # - m1 => feed 1, d1
+      # - m2 => feed 1, d2; feed 2, d3
+      # - m3 => feed 2, d1
+      # - m4 => feed 2, d2
+      # - m5 => feed 2, d4
+      #
+      # For metric 'm5', no metric data will be returned (simulating missing metric).
+      parser.collector = double("Mocked collector")
+      allow(parser.collector).to receive(:metrics_for_metric_type) do |feed, metric_type_id|
```
cfcosta: Where do we test this behaviour? There's lots of logic there, and it makes the test a lot harder to understand :(
I'd extract this mock into its own class and also test it.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
```diff
@@ -97,79 +97,149 @@ def inventory_object_data(inventory_object)
...
+                                    :hawkular_metric_id => 'm4')
+          metrics << OpenStruct.new(:path               => '/t;hawkular/f;2/r;s1/r;d3/m;m2',
+                                    :hawkular_metric_id => 'm2')
+          metrics << OpenStruct.new(:path               => '/t;hawkular/f;2/r;s1/r;d4/m;m5',
+                                    :hawkular_metric_id => 'm5')
+        end
+
+        metrics
+      end
+      allow(parser.collector).to receive(:raw_availability_data) do |ids|
```
cfcosta: Same.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
```diff
@@ -97,79 +97,149 @@ def inventory_object_data(inventory_object)
...
+
+  describe 'fetch_server_availabilities' do
+    it 'resolves availabilities correctly' do
+      parser = described_class.new
+
+      # Define a mocked persister.
+      # It has two feeds:
+      # - Feed 1 has one server with id: s1
+      # - Feed 2 has two servers ids: s1, s2
+      parser.persister = double("Mocked persister")
```
cfcosta: https://relishapp.com/rspec/rspec-mocks/docs/verifying-doubles

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
israel-hdez: Well... This is input data. If you change it, the output should change and expectations fail. Surely, I can extract it to a separate class (inside spec/support/ folder), but I see redundant to test it since it is like a fixture.

Given said that, I don't know if extracting the mocks is the better option. Extracting them will lead to have a few lines of code initializing the mocks followed by a call to `fetch_deployment_availabilities` and the expectations. That also won't allow to clearly understand the test.

I would wait for a third opinion here. If nobody comment, I will honor your suggestion.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
israel-hdez: Hmm... But if I extract the mocks to their own classes, will this apply? Or was you suggesting to extract the mocks to a helper method building the mocks?

cfcosta: What do you mean by "extract the mocks to their own clasess"? Every time you mock something you should at least use `instance_double` and point to the correct class so then the mocks can be verified if the interface changes.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
cfcosta: Not exactly. First I'd change so every test case tests one of the possibilities. The way you're doing it here is adding logic to the test that is not related to what the class is doing at all, but to the input, and the logic for the input should be tested on the class that is used for input. In that case, first I'd add a specific mock for each test case, and remove the logic from the parser.

Another option is to extract the mock behaviour into its own class, and also test that class.

israel-hdez: > First I'd change so every test case tests one of the possibilities.

I agree! This test is testing a lot. I will break it into the very specific cases.

> Another option is to extract the mock behaviour into its own class, and also test that class.

There are some cases where I still will need to add if/else logic to the mock. But wait until I broke this test into the specific test cases to see if the mock is acceptable. If not acceptable, then I'll place it separately.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
israel-hdez: Never mind... I was thinking in creating a mock-class and do something like:
`persister.parser` = MockedPersister.new
and I couldn't see how to use `instance_double`. But you were referring only to the behaviour in your other comment. That makes sense :)

israel-hdez:
@cfcosta Changed the specs to test each part and each behavior separately. I hope it is better now.

abonas:
@cfcosta please re-review so if it's ready I can merge asap

app/models/manageiq/providers/hawkular/middleware_manager/middleware_deployment.rb
```diff
@@ -1,4 +1,14 @@
 module ManageIQ::Providers
   class Hawkular::MiddlewareManager::MiddlewareDeployment < MiddlewareDeployment
+    def self.resource_path_for_metrics(item)
+      path = ::Hawkular::Inventory::CanonicalPath.parse(item.ems_ref)
```
cfcosta: What about extracting this whole method into a method in CanonicalPath? Doesn't even need to be on this PR, just a follow up with something like `parse_strict` or something like that.

israel-hdez: For this, CanonicalPath should be aware that the path is referring to a subdeployment. That shouldn't be problem. So, sounds good! Took note of this, but placing your suggestion in a github issue or an etherpad note may be better? That way, we could ask for team feedback before moving forward.

cfcosta: Yeah, sure, I've been writing some stuff down for the past few weeks, and plan to release as soon as I have something concrete. About this one, I think that's not exactly a problem, and can be changed ad-hoc.

lucasponce:
Sorry for my delay.
For some reasons the notification to review this PR was missing/filtered out in my inbox.
@Jiri-Kremser @jotak could you take a look as it is more to confirm that inventory changes will be ok ?
Thanks.

Jiri-Kremser:
@lucasponce looks good from the brief look, but it's hard to say without actually trying it. @israel-hdez has the MiQ working with @jotak's PR so I assume he can confirm.

israel-hdez:
@Jiri-Kremser It works for me! Although, this one is my PR. I don't know if it is valid to say that my own PR works.

abonas:
unless I misread the above - this PR fixes a bug and should not be relying on the new inventory (which might not be ready/merged in the same time)

israel-hdez:
@abonas yes, this PR fixes a bug and doesn't rely on the new inventory.

abonas:
> @abonas yes, this PR fixes a bug and doesn't rely on the new inventory.

ok, great. So I do not understand the above @Jiri-Kremser  and @lucasponce comments then regarding @jotak inventory changes and verifying them.

abonas:
is this ready to be merged?

lucasponce:
I am not updated in the latest changes on inventory and how it could affect into the hawkular API, and the use from ManageIQ that is why I put @jotak in the loop to keep an eye on it. 
I have no further info on this topic.
If everyone is happy, LGTM.

jotak:
Well, even if this one doesn't rely on new inventory, it's still worth checking that the other PR won't be impacted by this one... But as far as I can tell, it's fine :)

abonas:
@miq-bot add_label fine/yes

simaishi:
@israel-hdez Please resolve conflict and create an Euwe-specific PR (referencing this one) or suggest other PRs to backport.

Fine backport conflict:

```
$ git status
On branch fine
Your branch is up-to-date with 'origin/fine'.
You are currently cherry-picking commit 487a65c.
  (fix conflicts and run "git cherry-pick --continue")
  (use "git cherry-pick --abort" to cancel the cherry-pick operation)

Changes to be committed:

	modified:   app/models/manageiq/providers/hawkular/middleware_manager/middleware_deployment.rb
	modified:   spec/models/manageiq/providers/hawkular/middleware_manager/refresher_spec.rb
	modified:   spec/vcr_cassettes/manageiq/providers/hawkular/middleware_manager/refresher.yml

Unmerged paths:
  (use "git add/rm <file>..." as appropriate to mark resolution)

	deleted by us:   app/models/manageiq/providers/hawkular/inventory/collector/middleware_manager.rb
	both modified:   app/models/manageiq/providers/hawkular/middleware_manager/refresh_parser.rb
	both modified:   spec/models/manageiq/providers/hawkular/middleware_manager/refresh_parser_spec.rb
```

abonas:
given that this is medium priority/severity bug, and since it relies on a lot of code changes that are not in 'fine', and hence requires a complete rewrite of this bug fix just for 'fine', lets not backport it to 'fine'.
@miq-bot add_label fine/no
@miq-bot rm_label fine/yes

PR #9, opened 2017-04-14
Extract server operations from MiddlewareManager

cfcosta:
## What is it?

MiddlewareManager is, at the moment, not only the biggest class in the
project, but also the one with most churn[1], both of which are indicators
that the class is doing more than necessary.

This commit extract those methods to a concern, so that the interface
remains the same, but with a proper DSL on top to make the definitions
declarative instead of imperative.

[1]: For this PR, we consider "churn" the rate at which a file changes,
so more changes mean the file has more responsibilities, since objects
with loose coupling and high cohesion have less reason to change.

## Other

1. How to see what the churn for the files is?
    Run this:

        git log --all -M -C --name-only --format='format:' | sort | grep -v '^$' | uniq -c | sort -n | awk 'BEGIN {print "count\tfile"} {print $1 "\t" $2}' | grep hawkular | sort -nr | less


cfcosta:
Guys, I need some feedback on this, as it marks some of the ideas I have on extracting code and making it less dependant on MiQ, which will help us to move forward faster, change implementation without breaking stuff on MiQ and also identify places to be extracted as dependencies or moved back to hawkularclient.

@mtho11 @Jiri-Kremser @israel-hdez @abonas

lib/values/datasource.rb
```diff
@@ -0,0 +1,8 @@
+module ManageIQ
+  module Providers
+    module Hawkular
+      class Datasource
```
Jiri-Kremser: what's the purpose of this class?

app/models/manageiq/providers/hawkular/concerns/server_operations.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+
+      def remove_middleware_datasource(ems_ref)
+        run_generic_operation(operation_name, ems_ref)
+      end
+
+      def run_generic_operation(operation_name, ems_ref, params = {})
+        GenericServerOperation
+          .new(self, operation_name)
+          .run_generic(ems_ref, params)
```
Jiri-Kremser: perhaps I am blind, but where is the `run_generic` method defined? I can see only the `run` method in the `GenericServerOperation`

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -1,4 +1,27 @@
 describe ManageIQ::Providers::Hawkular::MiddlewareManager do
+  subject { described_class.new }
+
+  it "includes server operations" do
```
israel-hdez: Addittionally to the interface inclusion tests, I would test each one to ensure that the actual `run` is done. But I don't know if that is over-testing.

lib/generic_server_operation.rb
```diff
@@ -0,0 +1,41 @@
+module ManageIQ::Providers::Hawkular
+  class GenericServerOperation
```
israel-hdez: Now that you have this one here, I think that we should agree on what to place in MiQ vs in hawkular-client. For me, this one is candidate to be placed in hawkular-client., making it available to any user of the gem without reinventing the wheel.

spec/lib/middleware_notification_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    context 'class is MiddlewareServer' do
+    end
+
+    context 'entity found'
+    context 'entity not found'
+  end
+
+  describe '.emit_success'
+  describe '.emit_failure'
+  describe '#emit'
```
cfcosta: In standard method notation for Ruby, `Class#something` denotes an instance method, while `Class.something` is a class method or a singleton method.

app/models/manageiq/providers/hawkular/concerns/server_operations.rb
```diff
@@ -0,0 +1,87 @@
+module ManageIQ::Providers
+  module Hawkular
+    module ServerOperations
```
cfcosta: Actually, having concerns as a different folder, but without namespacing it is the rails default (try creating a new rails application and check the `app/models` folder.

lib/values/datasource.rb
cfcosta: Yes it is :)

cfcosta: Being removed.

app/models/manageiq/providers/hawkular/concerns/server_operations.rb
```diff
@@ -0,0 +1,282 @@
...
+      def suspend_middleware_server_group(ems_ref, params = {})
+        run_generic_operation('Suspend Servers', ems_ref, :timeout => params[:timeout] || 0)
+      end
+
+      def resume_middleware_server_group(ems_ref)
+        run_generic_operation('Resume Servers', ems_ref)
+      end
+
+      def create_jdr_report(ems_ref)
+        run_generic_operation(:JDR, ems_ref)
```
Jiri-Kremser: afaik, JDR reports are blocked by a flaw in the websocket library we are using. It was very likely it fails with seg fault when receiving the binary websocket frame. The probability was higher when the data was larger, so with JDR report it was almost sure, it fails. Once Ruben's [work](https://github.com/hawkular/hawkular-client-ruby/pull/205) is merged, this can be added.

abonas: @israel-hdez please see above JDR related code and info

abonas:
@cfcosta what's the status of this PR?

cfcosta:
@abonas I didn't work on it for a long time, will give it some love today.

cfcosta:
@abonas @israel-hdez @mtho11 this PR is ready, please review.

cfcosta:
This should be merged at the same time as ManageIQ/manageiq-ui-classic#2484.

app/models/manageiq/providers/hawkular/inventory/server_operations.rb
```diff
@@ -0,0 +1,278 @@
...
+            end
+
+            def event_message(entity)
+              attributes = {
+                :operation => operation_name,
+                :server    => entity.name,
+                :status    => type == :mw_op_success ? _('succeeded') : _('failed')
+              }
+
+              message = _('%{operation} operation for %{server} %{status}') % attributes
```
abonas: what is the `% attributes` part does here?

app/models/manageiq/providers/hawkular/inventory/server_operations.rb
```diff
@@ -0,0 +1,278 @@
...
+                :resourcePath         => ems_ref.to_s,
+                :datasourceName       => hash[:datasource]["datasourceName"],
+                :xaDatasource         => hash[:datasource]["xaDatasource"],
+                :jndiName             => hash[:datasource]["jndiName"],
+                :driverName           => hash[:datasource]["driverName"],
+                :driverClass          => hash[:datasource]["driverClass"],
+                :connectionUrl        => hash[:datasource]["connectionUrl"],
+                :userName             => hash[:datasource]["userName"],
+                :password             => hash[:datasource]["password"],
+                :xaDataSourceClass    => hash[:datasource]["driverClass"],
```
abonas: the driverClass is used both here and above in line 71 - is this a bug?

abonas:
reviewed. @israel-hdez please review as well

app/models/manageiq/providers/hawkular/inventory/server_operations.rb
cfcosta: If it is then it is wortking as such, and outside of the scope of this PR, given that I'm focusing on only refactoring and not changing anything else.

app/models/manageiq/providers/hawkular/inventory/server_operations.rb
cfcosta: Named string interpolation. It works like `printf`, `sprintf` and so on.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -28,6 +29,27 @@ class Hawkular::MiddlewareManager < ManageIQ::Providers::MiddlewareManager
...
+    domain_operation :kill, 'Kill'
+
+    group_operation :start, 'Start Servers'
+    group_operation :stop, 'Stop Servers'
+    group_operation :restart, 'Restart Servers'
+    group_operation :reload, 'Reload Servers'
+    group_operation :suspend, 'Suspend Servers'
+    group_operation :resume, 'Resume Servers'
+
+    generic_operation :create_jdr_report, 'JDR'
```
israel-hdez: You can remove this one. It's dead code.
JDR op is implemented in [middleware_diagnostic_report](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb) model.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -28,6 +29,27 @@ class Hawkular::MiddlewareManager < ManageIQ::Providers::MiddlewareManager
     has_many :middleware_messagings, :foreign_key => :ems_id, :dependent => :destroy
     has_many :middleware_server_groups, :through => :middleware_domains
 
+    standalone_operation :shutdown, 'Shutdown'
```
israel-hdez: Shouldn't this one (and some others) have `{:timeout => 0}` as default parameters?

mtho11: I believe the hawkular agent defaults this already, so no need to pass.

mtho11:
@abonas just curious if this code has to wait to be merged until after 4.6 because of the retesting QE would have to do? And they don't have the current resources to retest.

abonas:
> @abonas just curious if this code has to wait to be merged until after 4.6 because of the retesting QE would have to do? And they don't have the current resources to retest.

excellent question @mtho11 
I would not wait with merging this one for several reasons:
1. it is well before feature freeze
2. the refactoring here is mostly extracting existing code to different files without making changes to it, rather than rewriting logic/changing libraries that are used. so this is low risk refactoring compared to others.
3. server side change only, (and which is better covered by tests than the UI code).

gbaufake:
@cfcosta @mtho11 

I think It is possible to merge and test after it considering @abonas three reasons mentioned. Any problem with this PR, we may revert it or correct the wrong part.

I took a look on the code and it seems to everything is place.

Do we know which features may this pull request impact? If the answer is an yes, we could a "pre-merge" test as well.

Good Job @cfcosta!

mtho11:
Thanks for the feedback @gbaufake. I also think it looks great and would love to see it in before code freeze for support reasons.

abonas:
@cfcosta I would like to merge it, but want to make sure first the codeclimate issues are ok? (for some reason I cannot view their details atm)

abonas:
@cfcosta ?

cfcosta:
@abonas I checked the codeclimate issues, they are related to code I haven't refactored yet, and refactoring it is not worth, imho, because another layer of indirection is not worth the benefit.

To me, we can just merge it.

PR #10, opened 2017-04-17
Adaptations to have a working refresh for Hawkular's Inventory.next

israel-hdez:
Hawkular's inventory is being rewritten to store its inventory as metrics. The [hawkular-client-ruby](https://github.com/hawkular/hawkular-client-ruby) is being modified to make the inventory's change as transparent as possible for users of the ruby client, but this is not fully possible. This PR deals with changes required in MiQ side to work correctly with new's Hawkular inventory.

~~Note: This PR was worked on top of code changes of #6. To view the relevant differences, look only at the last commit.~~ #6 has been merged. Now, it this is rebased from master.

https://bugzilla.redhat.com/show_bug.cgi?id=1446286

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
@@ -180,18 +180,20 @@ module ManageIQ::Providers
               :feed_id      => status_metric_path.feed_id,
               :resource_ids => status_metric_path.resource_ids
             )
-            metric_id_by_resource_path[resource_path.to_s] = status_metric.hawkular_metric_id
+            metric_id_by_resource_path[URI.decode(resource_path.to_s)] = status_metric.hawkular_metric_id
           end
         end
+
         collection.each do |item|
           yield item, nil
 
-          path = item.model_class.try(:resource_path_for_metrics, item) || item.manager_uuid
+          path = URI.decode(item.model_class.try(:resource_path_for_metrics, item) || item.manager_uuid)
           next unless metric_id_by_resource_path.key? path
           metric_id = metric_id_by_resource_path[path]
           resources_by_metric_id[metric_id] = [] unless resources_by_metric_id.key? metric_id
           resources_by_metric_id[metric_id] << item
         end
+
         unless resources_by_metric_id.empty?
           availabilities = collector.raw_availability_data(resources_by_metric_id.keys,
                                                            :limit => 1, :order => 'DESC')
@@ -248,9 +250,10 @@ module ManageIQ::Providers
 
       def parse_messaging(messaging, inventory_object, config)
         parse_base_item(messaging, inventory_object)
-
         inventory_object.name = messaging.name
-        inventory_object.messaging_type = messaging.to_h['type']['name']
+
+        type_path = ::Hawkular::Inventory::CanonicalPath.parse(messaging.type_path)
+        inventory_object.messaging_type = URI.decode(type_path.resource_type_id)
 
         if !config.empty? && !config['value'].empty? && config['value'].respond_to?(:except)
           inventory_object.properties = config['value'].except('Username', 'Password')
--- a/spec/models/manageiq/providers/hawkular/middleware_manager/refresher_spec.rb
+++ b/spec/models/manageiq/providers/hawkular/middleware_manager/refresher_spec.rb
@@ -30,19 +30,20 @@ describe ManageIQ::Providers::Hawkular::MiddlewareManager::Refresher do
 
     # check whether the server was associated with the vm
     server = @ems_hawkular.middleware_servers.first
-    expect(server.lives_on_id).to eql(@vm.id)
-    expect(server.lives_on_type).to eql(@vm.type)
+    # TODO: Restore these expectations
+    # expect(server.lives_on_id).to eql(@vm.id)
+    # expect(server.lives_on_type).to eql(@vm.type)
     expect(@ems_hawkular.middleware_deployments).not_to be_empty
     expect(@ems_hawkular.middleware_datasources).not_to be_empty
     expect(@ems_hawkular.middleware_messagings).not_to be_empty
     expect(@ems_hawkular.middleware_deployments.first).to have_attributes(:status => 'Enabled')
     expect(@ems_hawkular.middleware_servers.first.properties).to have_attributes(
-      'Availability'            => 'unknown',
-      'Calculated Server State' => 'unknown'
+      'Availability'            => 'up',
+      'Calculated Server State' => 'running'
     )
     assert_specific_datasource(@ems_hawkular, 'Local~/subsystem=datasources/data-source=ExampleDS')
     assert_specific_datasource(@ems_hawkular,
-                               'Local~/host=master/server=s/subsystem=datasources/data-source=ExampleDS')
+                               'Local~/host=master/server=server-one/subsystem=datasources/data-source=ExampleDS')
     assert_specific_server_group(domain)
     assert_specific_domain_server
     assert_specific_domain
@@ -77,11 +78,11 @@ describe ManageIQ::Providers::Hawkular::MiddlewareManager::Refresher do
   end
 
   def assert_specific_server_group(domain)
-    server_group = domain.middleware_server_groups.find_by(:name => 'my-group')
+    server_group = domain.middleware_server_groups.find_by(:name => 'main-server-group')
     expect(server_group).to have_attributes(
-      :name     => 'my-group',
-      :nativeid => 'Local~/server-group=my-group',
-      :profile  => 'default',
+      :name     => 'main-server-group',
+      :nativeid => 'Local~/server-group=main-server-group',
+      :profile  => 'full',
     )
     expect(server_group.properties).not_to be_nil
     expect(server_group.middleware_deployments).to be_empty
@@ -89,12 +90,12 @@ describe ManageIQ::Providers::Hawkular::MiddlewareManager::Refresher do
   end
 
   def assert_specific_domain_server
-    server = @ems_hawkular.middleware_servers.find_by(:name => 's')
+    server = @ems_hawkular.middleware_servers.find_by(:name => 'server-one')
     expect(server).to have_attributes(
-      :name     => 's',
-      :nativeid => 'Local~/host=master/server=s',
+      :name     => 'server-one',
+      :nativeid => 'Local~/host=master/server=server-one',
       :product  => 'WildFly Full',
-      :hostname => '58ef2cf13071',
+      :hostname => 'ea174c22c065',
     )
     expect(server.properties).not_to be_nil
   end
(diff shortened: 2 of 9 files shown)


pilhuhn:
CC @abonas

abonas:
@israel-hdez thanks for this! is this necessary to rely here on PR #6?

abonas:
cc @moolitayer @simon3z

app/models/manageiq/providers/hawkular/inventory/collector/middleware_manager.rb
```diff
@@ -45,8 +45,11 @@ def config_data_for_resource(resource_path)
       connection.inventory.get_config_data_for_resource(resource_path)
     end
 
-    def metrics_for_metric_type(metric_path)
-      connection.inventory.list_metrics_for_metric_type(metric_path)
+    def metrics_for_metric_type(feed, metric_type_id)
```
cfcosta: I'd call this `metrics_for_type`, the second `metric` sounds redundant.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
```diff
@@ -97,79 +97,149 @@ def inventory_object_data(inventory_object)
...
+  describe 'fetch_deployment_availabilities' do
+    it 'resolves availabilities correctly' do
+      parser = described_class.new
+
+      # Define a mocked persister.
+      # It has two feeds:
+      # - Feed 1 has two deployments with ids: d1, d2
+      # - Feed 2 has three deployments with ids: d1, d2, d3, d4
+      parser.persister = double("Mocked persister")
+      deployments = []
```
cfcosta: You can have this as only one array declaration.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
```diff
@@ -97,79 +97,149 @@ def inventory_object_data(inventory_object)
...
+  describe 'fetch_server_availabilities' do
+    it 'resolves availabilities correctly' do
+      parser = described_class.new
+
+      # Define a mocked persister.
+      # It has two feeds:
+      # - Feed 1 has one server with id: s1
+      # - Feed 2 has two servers ids: s1, s2
+      parser.persister = double("Mocked persister")
+      servers = []
```
cfcosta: Same here, just one array declaration.

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -144,44 +144,64 @@ def fetch_server_entities
...
-          # Ensure consistency on keys (resource_path) used on metric_id_by_resource_path
-          path = ::Hawkular::Inventory::CanonicalPath.new(:tenant_id    => path.tenant_id,
-                                                          :feed_id      => path.feed_id,
-                                                          :resource_ids => path.resource_ids)
-          path = path.to_s
+
+        collection.each do |item|
+          yield item, nil
+
+          path = CGI.unescape(item.model_class.try(:resource_path_for_metrics, item) || item.manager_uuid)
```
cfcosta: One detail that might be important is that CGI is not exactly related (meaning-wise) to what the PR is doing, and I think `URI.unescape` or `URI.decode` would be better. There are some tokens that parse differently on both methods, but I think it is worth a try to make the code less all-over-the-place.

abonas:
@israel-hdez the above are the only changes needed accross all miq repos? no vcr cassettes need to be rerecorded? (tests of this PR pass so I wonder regarding cassettes)

israel-hdez:
@abonas 
> is this necessary to rely here on PR #6?

It depends on which PR gets merged first. If done separately, both PRs will have conflicting changes. I'm assuming that #6 will get merged first.

> no vcr cassettes need to be rerecorded?

Yes! VCR cassettes need to be recorded again. Tests are passing, because travis is using the latest released hawkular-client gem, which works against current inventory. I will need some help here to know what was the setup used when VCR cassettes were first recorded, to recreate it an rerecord everything.

app/models/manageiq/providers/hawkular/inventory/collector/middleware_manager.rb
israel-hdez: Right now, it sounds redundant but, if needed, we may add `metrics_for_resource_type` or `metrics_for_resource`, turning ambiguous a `metrics_for_type` method. Could `metrics_with_type` be a better name?

simon3z:
@abonas the official `hawkular-client` gem version bump will happen on `manageiq-gems-pending.gemspec`, right?

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
israel-hdez: Right. Since hawkular api is web-based, I'll change those `CGI` usages with `URI`.

israel-hdez:
@simon3z @abonas Is `hawkular-client` gem being used in another repo? Else, we may want to remove it from `manageiq-gems-pending.gemspec` and add it to `manageiq-providers-hawkular.gemspec`

abonas:
> @simon3z @abonas Is hawkular-client gem being used in another repo? Else, we may want to remove it from manageiq-gems-pending.gemspec and add it to manageiq-providers-hawkular.gemspec

@israel-hdez  yes, it is used also by the containers provider which code currently resides in the main manageiq repo.
And regarding gem location - I also recently raised the topic of the gem declaration location - that it should not be in the gems pending gemfile). I 100% agree it's better to move it where you suggested (in a separate PR). @Fryguy @durandom  ack?

Fryguy:
@abonas Yes.  The goal of manageiq-gems-pending is to not exist.  Everything in there should be ripped out into it's own repo or, if provider specific, into the provider repo.

durandom:
> it is used also by the containers provider which code currently resides in the main manageiq repo.

I suggest both, `openshift`, `kubernetes` and `hawkular` add it to their dependency in `.gemspec` then?!

moolitayer:
I expect the gem to be used in:
- manageiq-providers-kubernetes
- manageiq-providers-openshift
- manageiq-providers-hawkular
- manageiq-providers-datawarehouse (does not exist but I understood it might exist in the future)

> I suggest both, openshift, kubernetes and hawkular add it to their dependency in .gemspec then?!

if one repo bumps a version in a way that is not compatible with another (both cannot be satisfied)
wont we know about it in the main mangeiq repo only after the bump PR merge?

durandom:
> if one repo bumps a version in a way that is not compatible with another (both cannot be satisfied)
wont we know about it in the main mangeiq repo only after the bump PR merge?

yes, it's the same with any other gem dependency

Jiri-Kremser:
@israel-hdez I can't run the refresh successfully. 
I had to change the 
`/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb:257` to use the `nil?` instead of `.empty?` because `nil.empty?` was causing exceptions and
in `/app/models/manageiq/providers/hawkular/inventory/collector/middleware_manager.rb:52`
there is still a call to method that has to be removed. Do you have any newer version that works?

israel-hdez:
This is working fine with v2.9.0 gem. So, I will un-wip it and let to you the decision to merge before or at the same time of ManageIQ/manageiq-gems-pending#138 is merged. It should not be merged "after" gems-pending PR because the refresh won't work. (Although, if merged "before", travis will stay red until the gems-pending PR is merged).

@cfcosta You may wand to do a quick re-review in the meanwhile.
@miq-bot remove_label wip

israel-hdez:
@miq-bot assign abonas

spec/models/manageiq/providers/hawkular/middleware_manager/refresher_spec.rb
```diff
@@ -30,19 +30,20 @@
 
     # check whether the server was associated with the vm
     server = @ems_hawkular.middleware_servers.first
-    expect(server.lives_on_id).to eql(@vm.id)
-    expect(server.lives_on_type).to eql(@vm.type)
+    # TODO: Restore these expectations
```
israel-hdez: @Jiri-Kremser I will appreciate if you can instruct me how you did your setup for these expectations, so I can bring them back to life in a follow-up PR.

israel-hdez:
@miq-bot add_label fine/yes, blocker

Please see https://github.com/ManageIQ/manageiq-gems-pending/pull/138#issuecomment-297736552 for backport notes.

abonas:
clicking on code climate shows green status, so I'll merge

spec/models/manageiq/providers/hawkular/middleware_manager/refresher_spec.rb
Jiri-Kremser: They should appear if the hawkular-services are deployed on a virtual machine managed by RHEV provider and this provider is also added to MiQ. The cross-linking should just happen automatically. If I recall correctly the https://hawkular.torii.gva.redhat.com is running on RHEV and both the hawkular and RHEV can be added as providers to MiQ (juca can give you the access). Although, it will probably not work for the containerized hawular-services, because we no longer have the 

```bash
sudo echo $HOSTNAME > /etc/machine-id
```

..in the start script, because the docker image must run without the root on openshift.

@jpkrohling As the author, could you pls check that I am not lying here :] Also, do you have any idea how to workaround the ^ for containers? Perhaps we can change the permissions of `/etc/machine-id` so that it's writable by non-root guys, right? That would work. If you are not against it I am willing to add it back.

israel-hdez:
If this cannot be backported to `fine`, please merge ManageIQ#14927.

spec/models/manageiq/providers/hawkular/middleware_manager/refresher_spec.rb
jpkrohling: > They should appear if the hawkular-services are deployed on a virtual machine managed by RHEV provider and this provider is also added to MiQ. 

That's correct.

> The cross-linking should just happen automatically.

Depending on the definition of automatically, that's also correct :) @abonas can confirm how it works on the UI, but on this side, all we need to do is to set the `lives_on` field. Something else will compare these values and show that accordingly on the UI.

> If I recall correctly the https://hawkular.torii.gva.redhat.com is running on RHEV and both the hawkular and RHEV can be added as providers to MiQ (juca can give you the access).

Also correct.

> Also, do you have any idea how to workaround the ^ for containers? Perhaps we can change the permissions of /etc/machine-id so that it's writable by non-root guys, right? That would work. If you are not against it I am willing to add it back.

I have no idea about that, though, as it probably depends on the outcome of the discussions about the machine ID for containers. In my *opinion*, `/etc/machine-id` should not contain the hostname, or anything that doesn't conform with the spec that rules this file: https://www.freedesktop.org/software/systemd/man/machine-id.html

israel-hdez: @Jiri-Kremser @jpkrohling 
> If I recall correctly the https://hawkular.torii.gva.redhat.com is running on RHEV 

It outputs a "bad gateway" error to me :(

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq/pull/14927

Issue #8, opened 2017-04-12
move provider specific settings


durandom:
see the [amazon settings](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/config/settings.yml) for what can be moved

abonas:
@miq-bot assign @cfcosta

cfcosta:
This is merged already, can it be closed?

PR #14, opened 2017-04-25
Extract hawkular-related settings from miq-core

### LINKED ISSUE #13: move hawkular specific event names
similar to moving the settings we can also move event names (see [amazon](https://github.com/ManageIQ/manageiq-providers-amazon/pull/163/files) for an example)

cfcosta:
Fixes #13 and #8. Should be merged together with https://github.com/ManageIQ/manageiq/pull/14883.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -3,9 +3,41 @@
   :ems_hawkular:
     :event_handling:
       :event_groups:
+        :application:
+          :critical:
+          - hawkular_datasource.error
+          - hawkular_datasource_remove.error
+          - hawkular_deployment.error
+          - hawkular_deployment_remove.error
+          - hawkular_event.critical
+          :detail:
+          - hawkular_datasource.ok
+          - hawkular_datasource_remove.ok
+          - hawkular_deployment.ok
+          - hawkular_deployment_remove.ok
+          - hawkular_event
+:ems_refresh:
+  :hawkular:
+    :refresh_interval: 15.minutes
 :http_proxy:
   :hawkular:
     :host:
     :password:
     :port:
     :user:
+:log:
+  :level_datawarehouse: info
+  :level_datawarehouse_in_evm: error
+:product:
+  :datawarehouse_manager: false
+:workers:
+  :worker_base:
+    :event_catcher:
+      :event_catcher_hawkular:
+        :poll: 10.seconds
+      :event_catcher_hawkular_datawarehouse:
+        :alertable_tenants: ['_system', '_ops']
+        :poll: 1.minute
+    :queue_worker_base:
+      :ems_refresh_worker:
+        :ems_refresh_worker_hawkular: {}


cfcosta:
@miq-bot add_label enhancement,fine/no

moolitayer:
I listed some stuff in https://github.com/ManageIQ/manageiq/pull/14883 we may need to add here

config/settings.yml
```diff
@@ -3,9 +3,51 @@
   :ems_hawkular:
     :event_handling:
       :event_groups:
+
+:ems_refresh:
+  :hawkular:
+    :refresh_interval: 15.minutes
+
+:event_handling:
```
durandom: this needs to be nested under `ems->ems_hawkular` because merging settings will not merge arrays. It will merge hashes though.

config/settings.yml
```diff
@@ -3,9 +3,51 @@
...
+        :defaults:
+          :event_catcher_hawkular:
+            :poll: 10.seconds
+          :event_catcher_hawkular_datawarehouse:
+            :alertable_tenants: ['_system', '_ops']
+            :poll: 1.minute
+  :queue_worker_base:
+    :ems_metrics_collector_worker:
+      :defaults:
+        :ems_metrics_collector_worker_kubernetes:
```
durandom: like noted on the core PR, this should be moved into the kubernetes provider

config/settings.yml
```diff
@@ -3,9 +3,51 @@
...
     :host:
     :password:
     :port:
     :user:
+
+:workers:
+  :worker_base:
+    :defaults:
+      :event_catcher:
+        :defaults:
```
Jiri-Kremser: shouldn't these lines 36 and 38 be deleted? The original structure in the config [was](https://github.com/ManageIQ/manageiq/blob/82637f6960f3737f01ae990eb3288b44187b6abd/config/settings.yml#L1158:L1197):

```yaml
  :workers:
    :worker_base:
      :event_catcher:
        :event_catcher_hawkular:
```

Or am I missing something?

config/settings.yml
```diff
@@ -3,9 +3,51 @@
...
+            :alertable_tenants: ['_system', '_ops']
+            :poll: 1.minute
+  :queue_worker_base:
+    :ems_metrics_collector_worker:
+      :defaults:
+        :ems_metrics_collector_worker_kubernetes:
+          :metrics_port: 5000
+          :metrics_path: "/hawkular/metrics"
+  :ems_refresh_worker:
+    :defaults:
```
Jiri-Kremser: same here, is the `defaults` necessary?

config/settings.yml
durandom: indeed, the nesting under `:defaults` is wrong

cfcosta:
Updated both this PR and https://github.com/ManageIQ/manageiq/pull/14883 . Both should be correct now.

cfcosta:
@durandom can you check it again?

config/settings.yml
```diff
@@ -3,9 +3,43 @@
   :ems_hawkular:
     :event_handling:
       :event_groups:
+:ems_refresh:
+  :hawkular:
+    :refresh_interval: 15.minutes
+:event_handling:
```
durandom: this has to be under `:ems -> :ems_hawkular`
check the settings.yml in the [amazon provider](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/config/settings.yml#L4) for reference

cfcosta: @durandom can you check if is is correct now?

config/settings.yml
```diff
@@ -3,9 +3,43 @@
...
-  :hawkular:
-    :host:
-    :password:
-    :port:
-    :user:
+    :ems_refresh:
+      :hawkular:
+        :refresh_interval: 15.minutes
+    :event_handling:
+      :event_groups:
```
durandom: `:event_handling -> :event_groups` is already on line 4

config/settings.yml
```diff
@@ -3,9 +3,43 @@
...
+          - hawkular_deployment.error
+          - hawkular_deployment_remove.error
+          - hawkular_event.critical
+          :detail:
+          - hawkular_datasource.ok
+          - hawkular_datasource_remove.ok
+          - hawkular_deployment.ok
+          - hawkular_deployment_remove.ok
+          - hawkular_event
+    :http_proxy:
```
durandom: from here on this has to remain on top level.

cfcosta:
@durandom fixed!

durandom:
@Jiri-Kremser merge?

PR #15, opened 2017-04-27
Move dependency from manageiq-gems-pending

bdunne:
Add runtime dependency on hawkular-client.

### DIFF
--- a/manageiq-providers-hawkular.gemspec
+++ b/manageiq-providers-hawkular.gemspec
@@ -13,6 +13,8 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
+  s.add_runtime_dependency "hawkular-client", "~> 3.0.2"
+
   s.add_development_dependency "codeclimate-test-reporter", "~> 1.0.0"
   s.add_development_dependency "simplecov"
 end


abonas:
we need to see what is the right approach here as 2 providers depend on this gem, so placing it here requires discussion

bdunne:
@abonas What is the other provider?  It uses the hawkular-client gem directly?
Also, any idea why the tests failed with VCR issues?  It looks like master is green.

abonas:
@bdunne kubernetes/openshift

abonas:
cc @moolitayer

moolitayer:
cc @simon3z

moolitayer:
@bdunne as hawkular-client is being consumed by other components won't this make dependencies more complex?

bdunne:
@moolitayer It may, but there are better ways to simplify the dependency relationship.

bdunne:
@Jiri-Kremser It will fail to bundle in the `manageiq` repo if these two repos have incompatible requirements.  

However, I think the kubernetes provider should depend on the hawkuar provider and move the minimal amount of connection logic in the kubernetes repo to this repo.  Then the dependency on hawkular-client is only here.

moolitayer:
If I understand currently the plan is indeed to move all kubernetes and openshift code dealing with hawkular to this repo (under the Datawarehouse manager) but that will not happen anytime soon.

In the mean time I didn't understand why we are splitting the dependency.
We will potentially need the two teams to be aware of upgrades - in that case I would expect the shared code not to be in the gems.

@bdunne I'm sure you have a reason for doing this so can you please explain?

israel-hdez:
@Jiri-Kremser It is supposed to make things easier. Let's try, and refine if needed.

bdunne:
@moolitayer The dependency is currently on `manageiq-gems-pending` (a technical debt repo) and doesn't belong there since nothing in manageiq-gems-pending requires it.  This repo does require the gem and should list it as a dependency.

jrafanie:
> In the mean time I didn't understand why we are splitting the dependency.
> We will potentially need the two teams to be aware of upgrades - in that case I would expect the shared code not to be in the gems.

@moolitayer It's about being explicit about dependencies.  As we extract things, the consumer of various gem dependencies become harder to find, so those dependencies need to be moved to where they're used.  It might seem like more work to maintain gem dependencies in multiple places but it's better to be explicit than to have no idea who's using a gem. 

For example, if I need to bump a gem version for ruby 2.4 and it's specified in manageiq-gems-pending but not used there, it's unclear to me how I know if the new gem version works wherever in all places.  Do I bundle and grep through my entire ruby .gem directory to find any callers of the gem?  What if I grep wrong?

bdunne:
:turtle: bump
cc @Fryguy

PR #16, opened 2017-04-28
BZ 1446485 - Container linking does not work with EAP on Javaagent

Jiri-Kremser:
For EAPs using the java agent monitoring, the properties we are looking
for are not there. Therefore we also look for the agent resource in the
inventory where we should find the info about the immutability and if
it's running in the container.

https://bugzilla.redhat.com/show_bug.cgi?id=1446485

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
@@ -21,6 +21,20 @@ module ManageIQ::Providers
             server = persister.middleware_servers.find_or_build(eap.path)
             parse_middleware_server(eap, server)
 
+            # if immutable or in container flag is not set, check also the agent resource
+            if server.properties['Immutable'].nil? || server.properties['In Container'].nil?
+              agent_resource_id = 'Local%20JMX~org.hawkular:type%3dhawkular-javaagent'
+              feed_id = ::Hawkular::Inventory::CanonicalPath.parse(eap.path).feed_id
+              agent_resource_path = ::Hawkular::Inventory::CanonicalPath.new(:feed_id      => feed_id,
+                                                                             :resource_ids => [agent_resource_id])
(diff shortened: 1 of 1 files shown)


Jiri-Kremser:
@miq-bot add_label blocker, fine/backported
cc @pilhuhn

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -21,6 +21,20 @@ def fetch_middleware_servers
             server = persister.middleware_servers.find_or_build(eap.path)
             parse_middleware_server(eap, server)
 
+            # if immutable or in container flag is not set, check also the agent resource
```
cfcosta: I think could be on it's own method or even class.

Jiri-Kremser: I agree and the same holds for the [next 12 lines](https://github.com/Jiri-Kremser/manageiq-providers-hawkular/blob/b4627462ee5bf9824f5989f24c0ad95928991ee9/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L38:L50), but let's not complicate this as this should go in as fast as possible. We can do that in a followup PR.

Jiri-Kremser: a side note: it's probably the best to do the refactoring right before the branch for the release on MiQ is forked, because after that the back-porting becomes more difficult

pilhuhn:
LGTM, Approved!

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
cfcosta: Sure, I agree.

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq/pull/14941

PR #17, opened 2017-05-03
bump ruby to 2.3.3 and add 2.4.1

durandom:
cc @jrafanie 
same as https://github.com/ManageIQ/manageiq-providers-lenovo/pull/36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: ruby
 rvm:
-- '2.3.1'
+- '2.3.3'
+- '2.4.1'
 sudo: false
 cache: bundler
 env:


abonas:
@cfcosta please review
@Jiri-Kremser what about the hawkular ruby gem? does it need some work accompanying this change?

Jiri-Kremser:
@abonas our tests in hawkular ruby gem do not run on 2.4.1 because of some vcr issues (well the fakeweb sub-component). They do not support 2.4.1 atm. But the client itself seems to be ok, I smoked tested it via irb and worked well.

cfcosta:
The looks ok to me, but the tests fail on 2.4.1, both for the provider and the client gem. I'm checking them on 2.3.3, will report back in a few minutes.

abonas:
@Jiri-Kremser what is fakeweb? Do you mean webmock?
I think we can't leave the client on <2.4 for a long time and expect it to live well in a 2.4 env

cfcosta:
@abonas @Jiri-Kremser on 2.4.1, I'm getting vcr-related errors when running specs, is this related to fakeweb/webmock and ruby's version?

cfcosta:
Provider is ok on 2.3.3, has some failing tests related to VCR on 2.4.1, same for the client for me. LGTM.

Jiri-Kremser:
> @Jiri-Kremser what is fakeweb? Do you mean webmock?

Right, webmock.. updating it to ~> 3.0.0 helped a bit, but there are is still a lot of errors. Strange, it says it can't find the record in the cassette, even though it's there

jrafanie:
@Jiri-Kremser see https://github.com/abonas/kubeclient/pull/247 where @cben figured out how to get webmock 2.3.1 to work.  Note, [this HACK](https://github.com/ManageIQ/manageiq/pull/13104/commits/ff9b4675f35303d00f93132bc081328fdcfd7da9) gets things working with the existing webmock + ruby 2.4 but the solution @cben is better if you can work with him to figure out how to get this repo working on 2.4 with a newer webmock without monkey patching.  Thanks!

cfcosta:
@jrafanie @Jiri-Kremser imho we should move with the monkeypatch first, because it does not require us to change any tests. In a followup effort, we can add what @cben did, since we can have confidence that the tests are working by themselves and are not tied to webmock's implementation details.

Even better, we can change the tests after we move to 2.4.1, because by then we'll have fixed all technical debt on this.

jrafanie:
@cfcosta I would use the monkey patch to verify that ruby 2.4 works as is and if so, determine how much effort it is to upgrade webmock.  I agree that the patch can be used short term to get things working and we can do the webmock upgrade separately.

cben:
See also this comment:
https://github.com/ManageIQ/manageiq/pull/14270#issuecomment-298450537
in case you need to convert cassettes to webmock 2 format.

jrafanie:
I just upgraded our foreman api client gem for ruby 2.4 support [here](https://github.com/ManageIQ/foreman_api_client/pull/7) and had to follow @cben's suggestions to convert the cassettes (not-re-record, just migrate them using a script)

Jiri-Kremser:
@cfcosta this is now gtg, right?

PR #18, opened 2017-05-04
Don't attempt to query Hawkular if no metrics were provided

israel-hdez:
Hawkular returns with an error status code if an empty list of metrics
is sent in the query. Indeed, it doesn't make sense to query the
hawkular server if no metrics were requested and, instead, an empty
object can be returned.

One use case where this method is used, is when generating performance
reports for middleware servers. If there is one server that has never been
in running state or has been down for a "long" time, it may not have any
metrics (all them may have expired and purged). If this happens, the
"metrics" argument of collect_stats_metrics will be empty.

https://bugzilla.redhat.com/show_bug.cgi?id=1445702

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager/live_metrics_capture.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/live_metrics_capture.rb
@@ -54,6 +54,7 @@ module ManageIQ::Providers
     end
 
     def collect_stats_metrics(metrics, start_time, end_time, interval)
+      return [{}, {}] if metrics.empty?
       gauge_ids, counter_ids, avail_ids, metrics_ids_map = parse_metrics_ids(metrics)
       starts = start_time.to_i.in_milliseconds
       ends = (end_time + interval).to_i.in_milliseconds + 1


israel-hdez:
@miq-bot assign @mtho11 
@miq-bot add_label bug

I'm not sure if it should be fine/yes

Jiri-Kremser:
@lucasponce could you pls review?

@israel-hdez wrt fine/yes/no: the bz has no blocker on it, but it looks pretty serious to me and the fix is 1-liner so no big risk. Perhaps it's more question on @abonas or @pilhuhn if this should go into fine.

lucasponce:
LGTM perhaps this validation was missing at the beginning.

israel-hdez:
Thanks @lucasponce. Then, I guess this can be merged.

If this is going to be backported, then ManageIQ/manageiq-gems-pending#150 should also be backported because the BZ is only fully solved if both PRs are applied.
cc @abonas

abonas:
In order to understand if this is fine/yes or fine/no, I'd like to understand how often this can happen to a user.
You wrote above that this can happen if one server has no metrics - is this issue occurring when there is only one single server in the system? or does it happen also when there are several servers, and only one of them has no metrics?
Regardless, merging to master.

abonas:
@miq-bot add_label fine/no

israel-hdez:
For the error to happen, it's enough if one of several servers doesn't have metrics.

abonas:
ok thanks. we'll leave it as fine/no, it is not a regression and it is a corner case

PR #19, opened 2017-05-04
Exclude ManageIQ specs from RSpec

cfcosta:
This makes `bundle exec rspec spec` works as specified, and makes
startup of `rake spec` a little tiny bit faster.

### DIFF
--- a/.rspec_ci
+++ b/.rspec_ci
@@ -3,3 +3,4 @@
 --color
 --order random
 --profile 25
+--exclude-pattern "spec/manageiq/**/*_spec.rb"
--- a/lib/tasks_private/spec.rake
+++ b/lib/tasks_private/spec.rake
@@ -7,5 +7,4 @@ desc "Run all specs"
 RSpec::Core::RakeTask.new(:spec => ["app:test:initialize", "app:evm:compile_sti_loader"]) do |t|
   spec_dir = File.expand_path("../../spec", __dir__)
   EvmTestHelper.init_rspec_task(t, ['--require', File.join(spec_dir, 'spec_helper')])
-  t.pattern = FileList[spec_dir + '/**/*_spec.rb'].exclude(spec_dir + '/manageiq/**/*_spec.rb')
 end
--- a/.rspec
+++ b/.rspec
@@ -2,3 +2,4 @@
 --require spec_helper
 --color
 --order random
+--exclude-pattern "spec/manageiq/**/*_spec.rb"


cfcosta:
@miq-bot add_label refactoring

bin/setup
```diff
@@ -1,11 +1,17 @@
 #!/usr/bin/env ruby
 require 'pathname'
 
+def system(*args)
+  puts args.first
+  super
+end
```
Jiri-Kremser: Is this needed? With the calls bellow, it prints `git` and `find`, right?

Jiri-Kremser:
@josejulio could you pls look?

bin/setup
cfcosta: Nope, it prints the whole command.

Jiri-Kremser: ok, why not. Then, it's like `set -x` in bash and the scope is only for this script so we are not doing anything wild here.

cfcosta: Yeah, exactly, that's the idea, to at least know what we're running, instead of only the output of the command.

bin/setup
```diff
@@ -1,11 +1,17 @@
...
+
 gem_root = Pathname.new(__dir__).join("..")
 
 unless gem_root.join("spec/manageiq").exist?
   puts "== Cloning manageiq sample app =="
   system "git clone https://github.com/ManageIQ/manageiq.git --branch master --depth 1 spec/manageiq"
+  system "find spec/manageid -name '*_spec.rb' -delete"
```
josejulio: typo should be `spec/manageiq` instead of `spec/manageiD`

lib/tasks_private/spec.rake
```diff
@@ -7,5 +7,5 @@ desc "Run all specs"
 RSpec::Core::RakeTask.new(:spec => ["app:test:initialize", "app:evm:compile_sti_loader"]) do |t|
   spec_dir = File.expand_path("../../spec", __dir__)
   EvmTestHelper.init_rspec_task(t, ['--require', File.join(spec_dir, 'spec_helper')])
-  t.pattern = FileList[spec_dir + '/**/*_spec.rb'].exclude(spec_dir + '/manageiq/**/*_spec.rb')
+  t.pattern = FileList[spec_dir + '/**/*_spec.rb']
```
josejulio: This will run all the tests if specs/manageiq is a symlink (instead of a clone) see [bin/update](https://github.com/cfcosta/manageiq-providers-hawkular/blob/eff59769d1babb023ebf41ce86f0cb6f79bf1df4/bin/update#L6-L7).
We should verify if we are dealing with a symlink and update the `pattern`

bin/setup
josejulio: This might cause us troubles when calling `git pull` on [bin/update](https://github.com/cfcosta/manageiq-providers-hawkular/blob/eff59769d1babb023ebf41ce86f0cb6f79bf1df4/bin/update#L6-L7).
We maybe (I'm not exactly sure if there could be conflicts if the deleted files were modified) revert back those files (`git reset --hard` maybe?) previous to the `git pull` and redelete the specs (some newer specs might have been added).

abonas:
@durandom ols review this as you dealt with this area recently due to the repo split

durandom:
another shortcoming is for people with `spec/manageiq` symlinked to `<path_to_miq_clone>`
@cfcosta did you look into adding some options to `.rspec` ?  Maybe thats a viable root

cfcosta:
@bdunne @durandom Yeah, this is a better idea, changed.

durandom:
@cfcosta thanks for looking, did you check if we still need the exclusion in the rake task then? Maybe we can simplify that as well?

And would you mind to create a PR for the provider generator in core as well? The templates are at `lib/generators/providers`

bdunne:
@cfcosta Nice!  As @durandom mentioned, it looks like you can delete https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/lib/tasks_private/spec.rake#L10 as well, then this is ready to be merged.

lib/tasks_private/spec.rake
bdunne: I don't think this line is needed at all, isn't that the default pattern?

cfcosta: That's true, @bdunne !

cfcosta:
@bdunne @durandom done! Check https://github.com/ManageIQ/manageiq/pull/15039 :D

lib/tasks_private/spec.rake
```diff
@@ -4,8 +4,4 @@ namespace :spec do
 end
 
 desc "Run all specs"
-RSpec::Core::RakeTask.new(:spec => ["app:test:initialize", "app:evm:compile_sti_loader"]) do |t|
-  spec_dir = File.expand_path("../../spec", __dir__)
-  EvmTestHelper.init_rspec_task(t, ['--require', File.join(spec_dir, 'spec_helper')])
```
bdunne: This line wasn't supposed to be removed.  With this deletion we no longer use `.rspec_ci`

cfcosta: Fixed.

lib/tasks_private/spec.rake
```diff
@@ -5,7 +5,5 @@ end
 
 desc "Run all specs"
 RSpec::Core::RakeTask.new(:spec => ["app:test:initialize", "app:evm:compile_sti_loader"]) do |t|
-  spec_dir = File.expand_path("../../spec", __dir__)
   EvmTestHelper.init_rspec_task(t, ['--require', File.join(spec_dir, 'spec_helper')])
-  t.pattern = FileList[spec_dir + '/**/*_spec.rb'].exclude(spec_dir + '/manageiq/**/*_spec.rb')
```
durandom: I was thinking like this:

```ruby
RSpec::Core::RakeTask.new(:spec => ["app:test:initialize", "app:evm:compile_sti_loader"]) do |t|
  spec_dir = File.expand_path("../../spec", __dir__)
  EvmTestHelper.init_rspec_task(t, ['--require', File.join(spec_dir, 'spec_helper')])
  t.pattern = FileList[spec_dir + '/**/*_spec.rb']
end
```

but maybe there is even a simpler way.

cfcosta: @durandom I think that's good enough, there you go!

.rspec_ci
```diff
@@ -3,3 +3,4 @@
 --color
 --order random
 --profile 25
+--exclude-pattern "spec/manageiq/**/*_spec.rb"
```
durandom: :( travis doesnt seem to pick this up

lib/tasks_private/spec.rake
bdunne: @cfcosta Please see the changes in https://github.com/ManageIQ/manageiq-automation_engine/pull/23

cfcosta:
@durandom @bdunne done!

Jiri-Kremser:
> Current state has some issues, maybe we should reconsider if we really need/want this optimization.

@josejulio I guess, it's ok now, isn't it? If you are not against, I would merge

josejulio:
@Jiri-Kremser They are good, thanks!

cfcosta:
@abonas can this be merged?

PR #20, opened 2017-05-04
Unescape escaped characters from MiddlewareServer.feed

cfcosta:
BZ Ticket: https://bugzilla.redhat.com/show_bug.cgi?id=1446351

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server.rb
@@ -1,5 +1,9 @@
 module ManageIQ::Providers
   class Hawkular::MiddlewareManager::MiddlewareServer < MiddlewareServer
+    def feed
+      CGI.unescape(super)
+    end
+
     def immutable?
       properties['Immutable'] == 'true'
     end
--- a/app/models/manageiq/providers/hawkular/middleware_manager/alert_profile_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/alert_profile_manager.rb
@@ -100,10 +100,15 @@ module ManageIQ::Providers
     def calculate_member_data_id_map(server, group_trigger)
       data_id_map = {}
       prefix = group_trigger.context['dataId.hm.prefix'].nil? ? '' : group_trigger.context['dataId.hm.prefix']
+
+      feed = CGI.escape(server.feed)
+
       group_trigger.conditions.each do |condition|
-        data_id_map[condition.data_id] = "#{prefix}MI~R~[#{server.feed}/#{server.nativeid}]~MT~#{condition.data_id}"
+        id_prefix = "#{prefix}MI~R~[#{feed}/#{server.nativeid}]~MT~"
+
+        data_id_map[condition.data_id] = "#{id_prefix}#{condition.data_id}"
         unless condition.data2_id.nil?
-          data_id_map[condition.data2_id] = "#{prefix}MI~R~[#{server.feed}/#{server.nativeid}]~MT~#{condition.data2_id}"
+          data_id_map[condition.data2_id] = "#{id_prefix}#{condition.data2_id}"
         end
       end
       data_id_map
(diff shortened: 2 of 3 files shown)


cfcosta:
@abonas should this be backported?

abonas:
No, this is not a blocker and should not be backported

cfcosta:
@Jiri-Kremser @jmazzitelli @israel-hdez Yeah, I thought adding it on the textual helpers, but my reasoning was that CGI escaping was a transport matter and not inherent to the server feed. So the best idea in this case would be to always present it unescaped and escape if necessary, maybe `CanonicalPath` should be responsible for this.

israel-hdez:
@Jiri-Kremser Correct, there is no relationship.
@cfcosta Yesterday, I was thinking the same but broader: the ruby client is the one that should be handling all escaping/unescaping and apps should only see unescaped thing (except when getting the full CP string)

abonas:
[this test](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb#L76) might be making the wrong assumptions about escaping...

abonas:
one more thought - iiic in this PR it is escaped when retrieved from db. 
why isn't it also escaped in refresh parser prior to persisting it to db?

cfcosta:
@abonas I agree, it would be better to have it persisted unescaped and escape as necessary, but there are many places where this could be escaped or unescaped, and there's no reliable way to find it. As such, I think the best idea is to always return it unescaped and escape as necessary.

Jiri-Kremser:
> isn't the feed also residing on additional objects so it should be fixed in more places? (domain for instance)

Right, that's true.

In theory (if things are not escaped twice or even more times) the `unescape`  should be idempotent, so that we can call unescape on anything (possibly multiple times) and be sure, it's unescaped or call `unescape` followed by `escape` and be sure it's escaped just once. In the old inventory, sometimes the ids were escaped in the JSON payloads (when appearing in the canonical path) and sometimes not (`id: `). It also changed from time to time. I recall, the `=` were not encoded but then suddenly were. If the id can be part of the URL, which was the case of the old inventory, the http clients tend to escape those ids as well. This is why it's so messy. 

Actually the BZ require that we should lie about the feed id, because internally it [contains](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/spec/vcr_cassettes/manageiq/providers/hawkular/middleware_manager/refresher.yml#L94) the `%20`, well to be fair, in the standalone xml it was defined with the space, but internally it's stored with the `%20`.

cfcosta:
@Jiri-Kremser @abonas so, what do you think is the direction we should go? There are some pros and cons on my approach, the main pro being that we can trust that every time the feed is going to be unescaped, and we can escape it to our hearts content if necessary.

abonas:
@cfcosta we can go with your approach, I would however ask you to handle this in additional places where we know this also happens. just to save on getting bugs on another places too ;)

cfcosta:
@abonas updated with the only usage that I could find, there are two others, but they already depend on `hawk_escape_id`, which does the escaping.

cfcosta:
Fixed CodeClimate warning, hopefully.

abonas:
so the domain use case is covered ? (not by this PR)

cfcosta:
@abonas what do you mean? The domain use case is related to immutability, no? And in that case as well it is realted to the agent, not to miq.

abonas:
> @abonas what do you mean? The domain use case is related to immutability, no? And in that case as well it is realted to the agent, not to miq.

I meant the same property (feed) on domain - perhaps there's a similar issue too

app/models/manageiq/providers/hawkular/middleware_manager/middleware_server.rb
```diff
@@ -1,5 +1,9 @@
 module ManageIQ::Providers
   class Hawkular::MiddlewareManager::MiddlewareServer < MiddlewareServer
+    def feed
+      CGI.unescape(super)
```
himdel: @cfcosta This breaks with..

```
> o.feed
NoMethodError: undefined method `tr' for nil:NilClass
```

when feed is `nil`. (Breaking ui-classic's specs.)

Can you perhaps add some logic to only unescape when non-nil?

israel-hdez: @himdel  I'd say that the factory is wrong. No server is added by the provider without a feed. May be also the database column is wrongly defined and should be "not null".

PR #22, opened 2017-05-11
Avoid trigger id collisions when Hawkular is linked to multiple MiQ installations

israel-hdez:
This is a border case. Issue is discussed in ManageIQ/manageiq#10674.

@miq-bot add_label enhancement, alerts

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager/alert_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/alert_manager.rb
@@ -2,13 +2,15 @@ module ManageIQ::Providers
   class Hawkular::MiddlewareManager::AlertManager
     require 'hawkular/hawkular_client'
 
-    def initialize(alerts_client)
-      @alerts_client = alerts_client
+    def initialize(ems)
+      @ems = ems
+      @alerts_client = ems.alerts_client
     end
 
     def process_alert(operation, miq_alert)
-      group_trigger = convert_to_group_trigger(miq_alert)
+      group_trigger = convert_to_group_trigger(operation, miq_alert)
       group_conditions = convert_to_group_conditions(miq_alert)
+
       case operation
       when :new
         @alerts_client.create_group_trigger(group_trigger)
@@ -25,7 +27,44 @@ module ManageIQ::Providers
       end
     end
 
-    def convert_to_group_trigger(miq_alert)
+    def self.build_hawkular_trigger_id(ems:, alert:)
+      ems.miq_id_prefix("alert-#{extract_alert_id(alert)}")
+    end
+
+    def self.resolve_hawkular_trigger_id(ems:, alert:, alerts_client: nil)
+      alerts_client = ems.alerts_client unless alerts_client
+      trigger_id = build_hawkular_trigger_id(:ems => ems, :alert => alert)
+
+      if alerts_client.list_triggers([trigger_id]).blank?
+        trigger_id = "MiQ-#{extract_alert_id(alert)}"
+      end
+
+      trigger_id
+    end
+
+    private
+
+    def build_hawkular_trigger_id(alert)
+      self.class.build_hawkular_trigger_id(:ems => @ems, :alert => alert)
+    end
+
+    def resolve_hawkular_trigger_id(alert)
+      self.class.resolve_hawkular_trigger_id(:ems => @ems, :alert => alert, :alerts_client => @alerts_client)
+    end
+
+    def self.extract_alert_id(alert)
+      case alert
+      when Hash
+        alert[:id]
+      when Numeric
+        alert
+      else
+        alert.id
+      end
+    end
+    private_class_method :extract_alert_id
+
+    def convert_to_group_trigger(operation, miq_alert)
       eval_method = miq_alert[:conditions][:eval_method]
       firing_match = 'ALL'
       # Storing prefixes for Hawkular Metrics integration
@@ -37,7 +76,14 @@ module ManageIQ::Providers
       when "mw_accumulated_gc_duration"
         context = { 'dataId.hm.type' => 'counter', 'dataId.hm.prefix' => 'hm_c_' }
       end
-      ::Hawkular::Alerts::Trigger.new('id'          => "MiQ-#{miq_alert[:id]}",
+
+      trigger_id = if operation == :new
+                     build_hawkular_trigger_id(miq_alert)
+                   else
+                     resolve_hawkular_trigger_id(miq_alert)
+                   end
+
+      ::Hawkular::Alerts::Trigger.new('id'          => trigger_id,
                                       'name'        => miq_alert[:description],
                                       'description' => miq_alert[:description],
                                       'enabled'     => miq_alert[:enabled],
--- a/app/models/manageiq/providers/hawkular/middleware_manager/alert_profile_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/alert_profile_manager.rb
@@ -2,8 +2,9 @@ module ManageIQ::Providers
   class Hawkular::MiddlewareManager::AlertProfileManager
     require 'hawkular/hawkular_client'
 
-    def initialize(alerts_client)
-      @alerts_client = alerts_client
+    def initialize(ems)
+      @ems = ems
+      @alerts_client = ems.alerts_client
     end
 
     def process_alert_profile(operation, miq_alert_profile)
@@ -24,12 +25,10 @@ module ManageIQ::Providers
       unless old_assignments_ids.empty?
         to_remove_alerts_ids = old_alerts_ids - new_alerts_ids
         to_add_alerts_ids = new_alerts_ids - old_alerts_ids
-        to_remove_alerts_ids.each do |alert_id|
-          group_trigger = @alerts_client.get_single_trigger "MiQ-#{alert_id}", true
+        retrieve_hawkular_group_triggers(to_remove_alerts_ids).each do |group_trigger|
           unassign_members(group_trigger, profile_id, old_assignments_ids)
         end
-        to_add_alerts_ids.each do |alert_id|
-          group_trigger = @alerts_client.get_single_trigger "MiQ-#{alert_id}", true
+        retrieve_hawkular_group_triggers(to_add_alerts_ids).each do |group_trigger|
           assign_members(group_trigger, profile_id, old_assignments_ids)
         end
       end
@@ -39,8 +38,7 @@ module ManageIQ::Providers
       to_unassign_ids = old_assignments_ids - new_assignments_ids
       to_assign_ids = new_assignments_ids - old_assignments_ids
       if to_unassign_ids.any? || to_assign_ids.any?
-        old_alerts_ids.each do |alert_id|
-          group_trigger = @alerts_client.get_single_trigger "MiQ-#{alert_id}", true
+        retrieve_hawkular_group_triggers(old_alerts_ids).each do |group_trigger|
           unassign_members(group_trigger, profile_id, to_unassign_ids) unless to_unassign_ids.empty?
           assign_members(group_trigger, profile_id, to_assign_ids) unless to_assign_ids.empty?
         end
@@ -112,6 +110,16 @@ module ManageIQ::Providers
         end
       end
       data_id_map
+    end
+
+    private
+
+    def retrieve_hawkular_group_triggers(alert_ids)
+      alert_ids.map do |item|
+        trigger_id = ::ManageIQ::Providers::Hawkular::MiddlewareManager::
+          AlertManager.resolve_hawkular_trigger_id(:ems => @ems, :alert => item, :alerts_client => @alerts_client)
+        @alerts_client.get_single_trigger(trigger_id, true)
+      end
     end
   end
 end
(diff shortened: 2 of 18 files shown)


israel-hdez:
I think it's important to know if `fine` will get this code or if Hawkular provider is already "released/productized". Right now, `id`s are being generated on demand.

If it is fine/no, I will need to add a column in the database to store the id used when the trigger was created (store it when creating the trigger; read it when updating/deleting the trigger). Without the database column, MiQ will try to update/delete triggers using the new structure while Hawkular will have the old ids. If it's going to be fine/yes, the column is not needed.

@abonas  @lucasponce

abonas:
it is fine/no
@miq-bot add_label fine/no

israel-hdez:
Ok, then I will need to add the migrations I said.
@miq-bot add_label wip

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -6,4 +6,49 @@
...
     expect(described_class.description).to eq('Hawkular')
   end
+
+  describe "miq_id_prefix" do
+    let(:random_id) { SecureRandom.hex(10) }
+    let(:my_region) do
+      MiqRegion.my_region || FactoryGirl.create(:miq_region, :region => MiqRegion.my_region_number)
+    end
+    let(:random_region) do
+      region = Random.rand(1..99) while !region || region == my_region.region
```
cfcosta: Randomness in tests? That is a recipe for disasters (and false positives, not sure which is worse). Why exactly is this necessary? Tests should be deterministic.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -6,4 +6,49 @@
...
+
+    before(:each) do
+      # Ensure "my region" exists
+      my_region
+    end
+
+    it "must return non-empty string" do
+      rval = subject.miq_id_prefix
+
+      expect(rval).to be_a(String)
```
cfcosta: We don't test types in Ruby unless strictly necessary, and specially not on primitives.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -6,4 +6,49 @@
...
+      MiqRegion.my_region || FactoryGirl.create(:miq_region, :region => MiqRegion.my_region_number)
+    end
+    let(:random_region) do
+      region = Random.rand(1..99) while !region || region == my_region.region
+      MiqRegion.find_by(:region => region) || FactoryGirl.create(:miq_region, :region => region)
+    end
+
+    before(:each) do
+      # Ensure "my region" exists
+      my_region
```
cfcosta: You can use `let!(:my_region)` instead of doing this `before` hook. Check the `!` on `let!`.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -6,4 +6,49 @@
...
+
+      expect(rval).to be_a(String)
+      expect(rval.strip).not_to be_empty
+    end
+
+    it "must prefix the provided string/identifier" do
+      rval = subject.miq_id_prefix(random_id)
+
+      expect(rval).to end_with(random_id)
+      expect(rval).not_to be == random_id
```
cfcosta: `expect(rval).not_to eq random_id`

spec/models/manageiq/providers/hawkular/middleware_manager/alert_profile_manager_spec.rb
```diff
@@ -1,6 +1,12 @@
 describe ManageIQ::Providers::Hawkular::MiddlewareManager::AlertProfileManager do
   let(:client) { double('Hawkular::Alerts') }
-  let(:subject) { described_class.new(client) }
+  let(:stubbed_ems) do
+    ems = instance_double('::ManageIQ::Providers::Hawkular::MiddlewareManager',
```
cfcosta: This smells like `ask, don't tell` to me. Maybe you should either receive the id_prefix as an argument, or move the code related to the usage of this to MiddlewareManager itself.

app/models/manageiq/providers/hawkular/middleware_manager/alert_profile_manager.rb
```diff
@@ -25,11 +26,11 @@ def update_alerts(profile_id, old_alerts_ids, new_alerts_ids, old_assignments_id
         to_remove_alerts_ids = old_alerts_ids - new_alerts_ids
         to_add_alerts_ids = new_alerts_ids - old_alerts_ids
         to_remove_alerts_ids.each do |alert_id|
-          group_trigger = @alerts_client.get_single_trigger "MiQ-#{alert_id}", true
+          group_trigger = @alerts_client.get_single_trigger(@ems.miq_id_prefix("alert-#{alert_id}"), true)
```
cfcosta: Why not move the `miq_id_prefix` call to `get_single_trigger`? Either that or moving it to a different class altogether. Having it here, and specially duplicated later, generate pain points and [Divergent Changes](https://sourcemaking.com/refactoring/smells/divergent-change), making it easier to generate bugs if we change one but not the other.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -6,4 +6,49 @@
...
+      ems_a = FactoryGirl.create(:ems_hawkular)
+      ems_b = FactoryGirl.create(:ems_hawkular)
+
+      expect(ems_a.miq_id_prefix).not_to be == ems_b.miq_id_prefix
+    end
+
+    it "must generate different prefixes for same provider on different MiQ region" do
+      ems_a = FactoryGirl.create(:ems_hawkular)
+      ems_b = ems_a.dup
+      ems_b.id = described_class.id_in_region(ems_a.id % described_class::DEFAULT_RAILS_SEQUENCE_FACTOR, random_region.region)
```
cfcosta: I don't understand what this is supposed to do. Does not sound very descriptive, and is kinda cryptic to use.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
israel-hdez: In this context, I'm testing `miq_id_prefix` which for any two instances of `MiddlewareManager` should comply with:
* It must return the same prefix if they live in the same region and they are pointing to the same provider (i.e. they come from the same database record, or they are two records from two different databases but are logically the same record).
* It must return different prefixes if they live in different regions, regardless if they point to the same provider.
* It must return different prefixes if they live in the same region, but they point to different providers.

These are the requirements of the generated prefix. But there is also the requirement that the prefix should be generated only from the data of one provider (you shouldn't pass the other provider as an argument) and the method should return something that must follow those rules.

I only need a `MiqRegion` that it's different to what `MiqRegion.my_region` returns. I don't know what `MiqRegion.my_region` will return, so I just randomly chose something else.

I'm not sure how this can lead to false positives since it is enough to have two different regions to have the tests do its work. In a broader view, I think the tests are kind of deterministic.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
israel-hdez: By changing the `id` I'm changing its region. Then, I have two instances of `MiddlewareManager` pointing to the same provider, but are in two different regions. 

Yes, I also don't like it. May be adding `#comment` to explain it can be enough? Adding a method to `MiqRegion` won't be useful since changing the region of a record is something that won't happen in real life scenarios.

spec/migrations/20170523220843_add_hawkular_id_to_miq_alerts.rb
```diff
@@ -0,0 +1,16 @@
+require_migration
+
+describe AddHawkularIdToMiqAlerts do
```
israel-hdez: How can I make this work?
Seems invisible to the main app, and there is no rake task in the local to run this spec.

israel-hdez:
@miq-bot remove_label wip

spec/migrations/20170523220843_add_hawkular_id_to_miq_alerts.rb
israel-hdez: @durandom

israel-hdez:
Hopefully, this PR is ready. Reviewers needed :mag:

spec/migrations/20170523220843_add_hawkular_id_to_miq_alerts.rb
durandom: @israel-hdez migrations have to go to the core app. Soon ™️ there will be a migrations repo for that

app/models/manageiq/providers/hawkular/middleware_manager/alert_manager.rb
```diff
@@ -37,7 +42,14 @@ def convert_to_group_trigger(miq_alert)
       when "mw_accumulated_gc_duration"
         context = { 'dataId.hm.type' => 'counter', 'dataId.hm.prefix' => 'hm_c_' }
       end
-      ::Hawkular::Alerts::Trigger.new('id'          => "MiQ-#{miq_alert[:id]}",
+
+      trigger_id = if operation == :new
+                     @ems.miq_id_prefix("alert-#{miq_alert[:id]}")
```
lucasponce: So, then triggers created in Hawkular Alerting will start with MiQ-region-xyz-ems-abc-alert-id. am I right ?
I know that create an integration test from MiQ is not trivial, but all the flow could be validated ?

spec/migrations/20170523220843_add_hawkular_id_to_miq_alerts.rb
israel-hdez: @durandom Ugh!.... I didn't see that answer coming. Ok, I'll move them there.

app/models/manageiq/providers/hawkular/middleware_manager/alert_manager.rb
israel-hdez: Yes, that's how IDs will look.
And yes, integration tests aren't trivial. But it may not be too hard to do what you are asking for. I'll try.

lucasponce:
For me the patch looks overall good.
I would make some live tests against Hawkular and check that alerts created from MiQ are stored in Hawkular as expected, and all the flow of the reception of events is working.

cfcosta:
@lucasponce I disagree we need to do live tests against Hawkular. That'd mean we're testing hawkular-client-ruby behaviour on the provider, which should not be the case.

lucasponce:
This particular integration is hard. 
As when you create an MiQAlert this should be created on the backend.
Also with the alert profiles.
But a integration test is needed, to reflect that changes are propagated correctly between MiQ and provider backend.
Propably it can be recorded, but I am not sure, that my suggestion to at least validate in a live test this intergation.

israel-hdez:
@lucasponce I added the integration tests. They are "flows", so only 5 samples that do multiple checks through each flow. I recorded the tests. Because of this I had to place fixed ids in the database and there is a rubocop warning as a consequence. I really don't like the idea to use fixed ids, but I had no choice, else vcr couldn't match the requests. Live test would just pass without fixed ids.

There are some other recorded tests that may be benefited if are converted to live tests. So, I will take the task to slowly migrate those test to true live tests to finally remove those cassettes.

@cfcosta After adding integration tests, they makes sense for me. It isn't only verifying that hawkular got called, but also the interaction between the main manager, the alert manager, the profiles manager, the database entities and, at the end, expecting the triggers got created in Hawkular. But, please comment if you see something wrong.

israel-hdez:
By the way, travis is red because I moved the migrations to the main repo. Once ManageIQ/manageiq#15306 is merged I can trigger another travis build to see if it gets green (locally, tests pass).

Would you mind to review again and comment if those tests are enough?

mtho11:
@israel-hdez exactly who do you want to review again?

israel-hdez:
@mtho11 it's just to know if the added tests are OK.

lucasponce:
@israel-hdez the tests looks good thanks. I haven't tested locally but they test the flow which is the complex part as there is a shared state between MiQ and Hawkular so some tests are you propose are necessary if we want to test this.

israel-hdez:
@mtho11 Now, there is no need of new database columns. I hope is ready to go (unless any new comments arise)

abonas:
@lucasponce @mtho11 id this is ready, can it be merged? if not, what's missing for it to be ready? thanks

lucasponce:
This PR adds an ems.region/guid identifier into the trigger ids.
As trigger ids are stored in the backend it protects against the collision in the corner case that a backend is managed by different miq instances.
Also, it deals with existing data, old identifiers are managed different than new ones to avoid problems in migrations.
There are new tests to validate the integration flow (which is not trivial, as it requires a dialog and state interchange between miq and provider).
I haven't tested locally but I understand the changes and flow.
Thx @israel-hdez for the work.
I would put a note on QA to include this change into their test suites, as it is not trivial and corner cases could appear.

From my side I don't have more comments.

abonas:
@mtho11 anything preventing this from being merged?

mtho11:
@abonas yes, this should be merged.

PR #24, opened 2017-06-05
[WIP] Remove the hawkular datawarehouse manager

moolitayer:
Moving away from using the datawarehouse manager as an external provider to integrating it in container providers. This will also ease the rename of the datawarehouse manager[1]

Note: all feature related to the Datawarehouse manager are in prototype mode

[1] https://github.com/ManageIQ/manageiq/pull/15295


abonas:
@moolitayer please suggest who should be reviewing this from cm

moolitayer:
@simon3z Can you please review?
The reason I'm doing this at this step is that i'd like to avoid adding another datawarehouse manager and add the new one straight as a monitoring manager. So: remove this, then rename the manager([here](ManageIQ/manageiq#15295)), then add the new manager

abonas:
@Jiri-Kremser @jshaughn @lucasponce  is this influencing our code in any way?

moolitayer:
@yakkov @zgalor please review this change

Jiri-Kremser:
@abonas It should not, I was grepping the code-base for the usage and didn't find anything, but Ruby can be tricky with its dynamicity

the spec errors look odd (https://travis-ci.org/ManageIQ/manageiq-providers-hawkular/jobs/240422563#L826)

@moolitayer is there a corresponding PR in the UI repo? I recall you had the nice general metric browser component, how is this related to this?

btw. nice [easter egg](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/24/files#diff-452a290fec66c0b661ece3f5305c5f69L27) :D

zgalor:
LGTM.
I added a comment in https://github.com/ManageIQ/manageiq/pull/15295 regarding manageiq-ansible-module dependency on this.

moolitayer:
> @moolitayer is there a corresponding PR in the UI repo? I recall you had the nice general metric browser component, how is this related to this?

Yes, removing the DWH screens is done in https://github.com/ManageIQ/manageiq-ui-classic/pull/1502. The ad hoc metric will remain as it is under the container manager. The end goal is to have it under the container provider's monitoring manager. Let me know if more info is needed.

moolitayer:
@yakkov please take a look

moolitayer:
@yaacov please review

yaacov:
LGTM because we haven't started using this provider ...

p.s.
When was this discussed ? I think we do need monitoring providers, IMO they are a good way to abstract the different monitoring endpoints.

moolitayer:
We are not removing the monitoring manager, only the hawkular implementation. We will end up having a monitoring manager under our provider (as well as a container manager).

jshaughn:
@abonas I don't think this affects our provider code.  If I understand, @moolitayer was building a generic "datawarehouse/monitoring" provider.  The code being removed was code from that effort, and would have been code eventually shared in the MW and container providers, and maybe others at some point.  This was when the container provider was targeting hawkular services/metrics.  Now that it is targeting prometheus I think Mooli has decided to pull back this effort and concentrate on just getting the container provider working.  I may not have this totally right.  Also, it's not clear whether, in a similar way, there will potentially be shareable code for p8s backed providers as part of the container effort. Mooli, can you let us know if what I'm saying is basically correct?

moolitayer:
Yes it is correct @jshaughn 
- The shared code should probably remain here(although not being currently used by middleware) I will change that.
- Things are still changing. I think I will leave the renames dependent on this and do my changes in a non intrusive way. Once I get the final decision on our direction I will move out of WIP

@miq-bot add_label wip

simon3z:
Approving because I am OK with the idea of removing the datawarehouse provider, @cben can help with the review of the code.

cben:
code changes are ✂️🔥✂️🔥 sounds good to me :-)
seriously though, is there any specific aspect I should review?

israel-hdez:
@moolitayer Will this PR move forward? Looks abandoned, and also all related PRs look the same.

moolitayer:
@israel-hdez I will get back to working on this as soon as I can

moolitayer:
Closing, see https://github.com/ManageIQ/manageiq/pull/16834#issuecomment-358675878

PR #25, opened 2017-06-06
Fix power operations 500 error on offline servers

cfcosta:
Bugzilla Ticket: https://bugzilla.redhat.com/show_bug.cgi?id=1444940


cfcosta:
@miq-bot add_label bug

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -508,11 +508,20 @@ def run_operation(parameters, operation_name = nil)
...
+
+        begin
+          operation_connection = connection.operations(true)
+
+          if operation_name.nil?
+            operation_connection.invoke_generic_operation(parameters, &callback)
+          else
+            operation_connection.invoke_specific_operation(parameters, operation_name, &callback)
+          end
+        rescue Errno::ECONNREFUSED => e
```
abonas: perhaps the rescue should be expanded to additional errors. not only connection refused. if hawkular server is up and returns 500 err, for example, imo it can behave the same on miq side

cfcosta:
@mtho11 will do!
@abonas should I catch all errors on this or some specifics?

abonas:
> @abonas should I catch all errors on this or some specifics?

we should present errors to the user in a clear way (and not "crash" MIQ while doing that) - so it sounds like yes, we should catch all errors in such case.

cfcosta:
@israel-hdez that's completely true, I'm going to move the change to hawkular client.

cfcosta:
Closed in favour of https://github.com/hawkular/hawkular-client-ruby/pull/223.

PR #26, opened 2017-06-07
Client error name format unification

cfcosta:
Depends on https://github.com/hawkular/hawkular-client-ruby/pull/222,
which adds new names for the errors on Hawkular ruby client.

### DIFF
--- a/manageiq-providers-hawkular.gemspec
+++ b/manageiq-providers-hawkular.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "hawkular-client", "~> 3.0.2"
+  s.add_runtime_dependency "hawkular-client", "~> 4.1"
 
   s.add_development_dependency "codeclimate-test-reporter", "~> 1.0.0"
   s.add_development_dependency "simplecov"
--- a/app/models/manageiq/providers/hawkular/datawarehouse_manager.rb
+++ b/app/models/manageiq/providers/hawkular/datawarehouse_manager.rb
@@ -19,9 +19,9 @@ module ManageIQ::Providers
       connect(options).fetch_version_and_status
     rescue URI::InvalidComponentError
       raise MiqException::MiqHostError, "Host '#{hostname}' is invalid"
-    rescue ::Hawkular::BaseClient::HawkularConnectionException
+    rescue ::Hawkular::ConnectionException
       raise MiqException::MiqUnreachableError, "Unable to connect to #{hostname}:#{port}"
-    rescue ::Hawkular::BaseClient::HawkularException => he
+    rescue ::Hawkular::Exception => he
       raise MiqException::MiqInvalidCredentialsError, 'Invalid credentials' if he.status_code == 401
       raise MiqException::MiqHostError, 'Hawkular not found on host' if he.status_code == 404
       raise MiqException::MiqCommunicationsError, he.message
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -37,9 +37,9 @@ module ManageIQ::Providers
         connect(options).inventory.list_feeds
       rescue URI::InvalidComponentError
         raise MiqException::MiqHostError, "Host '#{hostname}' is invalid"
-      rescue ::Hawkular::BaseClient::HawkularConnectionException
+      rescue ::Hawkular::ConnectionException
         raise MiqException::MiqUnreachableError, "Unable to connect to #{hostname}:#{port}"
-      rescue ::Hawkular::BaseClient::HawkularException => he
+      rescue ::Hawkular::Exception => he
         raise MiqException::MiqInvalidCredentialsError, 'Invalid credentials' if he.status_code == 401
         raise MiqException::MiqHostError, 'Hawkular not found on host' if he.status_code == 404
         raise MiqException::MiqCommunicationsError, he.message


Jiri-Kremser:
lgtm, we can merge once it's merged in the hawkular client
btw. there might be a small conflict with https://github.com/ManageIQ/manageiq-providers-hawkular/pull/24 luckily, removing the change from the datawarehouse provider will fix the codeclimate thing

cfcosta:
@Jiri-Kremser I updated the client version for the new version, so we have the new errors now.

manageiq-providers-hawkular.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "hawkular-client", "~> 3.0.2"
+  s.add_runtime_dependency "hawkular-client", "~> 4.0.0"
```
Fryguy: Since this is a shared dependency with manageiq-providers-kubernetes (and thus, manageiq-providers-openshift as well), please verify that a major bump will not break them.  https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/e0e4b5f427aec60e28a15cafd07345f48fcdee25/manageiq-providers-kubernetes.gemspec#L16

abonas: adding @moolitayer  - is this dependency on hawkular client still valid on k8s/ose providers considering the move to P8s?

moolitayer: Yes we will retain the existing metrics functionality for now.
`~> 4.0.0` and `~> 3.0.2` cannot both be satisfied

Fryguy: > P8s

Long name please? 😄

moolitayer: Prometheus like k8s for kubernetes replacing 8 letter

cfcosta: Created this PR to also change version number on manageiq-providers-kubernetes: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/56

israel-hdez:
@cfcosta is this going to be merged soon? I've some changes on top of 4.0.0... So, I'll need this soon (or backport my changes to 3.0.x).

cfcosta:
@israel-hdez add your changes and release a new version, I'll add it to this PR, merge it and ping https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/56 for them to also do the update.

israel-hdez:
@cfcosta My changes are in v4.1.0.
Please, update your PR... I'm not sure if we should keep using `~> x.x.x`. Every time the gem is released, we will need to update two repos.

abonas:
> @cfcosta My changes are in v4.1.0.
> Please, update your PR... I'm not sure if we should keep using ~> x.x.x. Every time the gem is released, we will need to update two repos.

@israel-hdez which 2 repos? and what other versioning do you propose? we use the standard semantic versioning approach so far.

cfcosta:
@israel-hdez if Gemfile.lock is not committed, I think we can move to `~> 4.0`, so we get nonbreaking changes on both.

@abonas he meant providers-hawkular and providers-kubernetes, the providers that currently are using the gem, and his suggestion is to use `gem 'hawkularclient', '~> 4.0'` instead of `4.0.0`, which only matches for minor changes. Given that we are taking a lot of care to mark all breaking changes, I think this is the way to go.

abonas:
> @israel-hdez if Gemfile.lock is not committed, I think we can move to ~> 4.0, so we get nonbreaking changes on both.
> 
> @abonas he meant providers-hawkular and providers-kubernetes, the providers that currently are using the gem, and his suggestion is to use gem 'hawkularclient', '~> 4.0' instead of 4.0.0, which only matches for minor changes. Given that we are taking a lot of care to mark all breaking changes, I think this is the way to go.

afaik kubernetes should drop hawkular client dependency soon.  @moolitayer @simon3z is that correct?

@israel-hdez @cfcosta what are the implications of moving from 3 digits versioning to 2 digits, and what is the policy with other gems in miq for versioning?

cfcosta:
@abonas the implications are that `4.0.0` matches anything from `4.0.0` to `4.0.99`, while `4.0` matches anything from `4.0.0` to `4.99.99`. Currently, there's no policy for that set in place on ManageIQ, some gems are locked on the patch version, some are locked on the minor version and some do not lock version at all, from what I've seen mostly test-based gems.

About Kubernetes, probably, but the PR is already there and currently they are still using it: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/56

abonas:
@cfcosta @israel-hdez ok. I agree we can go with 2 digit dependency here (but keep releasing the gem as a 3 digit recommended sem ver policy)
cc @mtho11 @karelhala @pilhuhn

cfcosta:
@abonas yeah, I don't think it is even possible to release versions containing only major and minor.

mtho11:
Merging so providers-kubernetes can use this dependency.

mtho11:
@cfcosta @abonas is this a `fine/yes` label?

cfcosta:
@mtho11 not sure, but I think so.

abonas:
@mtho11 fine has been released some time ago. why this should be fine/yes?

mtho11:
@abonas I was asking if it needs to be backported to fine as well?

abonas:
@mtho11 I don't see any reason to backport it but perhaps I'm missing something, why do you think it should be backported?

mtho11:
@abonas not unless @cfcosta has reasons.

cfcosta:
Not from me.

simon3z:
> afaik kubernetes should drop hawkular client dependency soon. @moolitayer @simon3z is that correct?

@abonas we won't drop it for a while, we'll support older OpenShift deployments for another year or so (exactly for how long depends on OpenShift release schedule).

PR #27, opened 2017-06-14
Extending event catcher and adding targeted refresh for availabilities

israel-hdez:
* A targeted refresh is added.
  * This targeted refresh only updates the status field of deployment entities when an availability update is received by the refresh worker.
  * Targeted refresh also updates server entities with the fetched new status.
  * No other field is updated, nor entity creation or deletion is performed. That work is left to the full graph refresh.
* Event catcher is extended to retrieve availability metrics from Hawkular for all deployments in inventory. Availabilities are "parsed" and compared against the status stored in MiQ database. Deployments whose status has changed are enqueued to be updated in the refresh worker.
* Event catcher is extended to fetch availability metrics of each inventoried server in MiQ. Also, its server state is fetched from Hawkular's inventory. A targeted refresh is queued for servers that have changed their availability or status.

With these changes, status of deployments and servers will be updated very frequently avoiding any manual refresh.

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/availability_updates.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/availability_updates.rb
@@ -1,0 +1,17 @@
+module ManageIQ::Providers
+  class Hawkular::Inventory::AvailabilityUpdates
+    delegate :select, :to => :@targets
+
+    def initialize(targets)
+      @targets = targets
+    end
+
+    def name
+      "Collection of availabilities to update on inventory entities"
+    end
+
+    def id
+      "Collection: #{@targets.map(&:manager_ref)}"
+    end
+  end
+end
--- a/app/models/manageiq/providers/hawkular/inventory/availability_updates.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/availability_updates.rb
@@ -1,6 +1,9 @@
 module ManageIQ::Providers
   class Hawkular::Inventory::AvailabilityUpdates
-    delegate :select, :to => :@targets
+    attr_reader :targets
+
+    delegate :select, :to => :targets
+    delegate :<<, :to => :targets
 
     def initialize(targets)
       @targets = targets
--- a/app/models/manageiq/providers/hawkular/inventory/collector/availability_updates.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/collector/availability_updates.rb
@@ -1,0 +1,11 @@
+module ManageIQ::Providers
+  class Hawkular::Inventory::Collector::AvailabilityUpdates < ManagerRefresh::Inventory::Collector
+    def deployment_updates
+      @target.select { |item| item.association == :middleware_deployments }
+    end
+
+    def server_updates
+      @target.select { |item| item.association == :middleware_servers }
+    end
+  end
+end
--- a/app/models/manageiq/providers/hawkular/inventory/parser/availability_updates.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/parser/availability_updates.rb
@@ -1,7 +1,15 @@
 module ManageIQ::Providers
   class Hawkular::Inventory::Parser::AvailabilityUpdates < ManagerRefresh::Inventory::Parser
     def parse
+      fetch_server_availabilities
       fetch_deployment_availabilities
+    end
+
+    def fetch_server_availabilities
+      collector.server_updates.each do |item|
+        server = persister.middleware_servers.find_or_build(item.manager_ref[:ems_ref])
+        server.properties = item.options
+      end
     end
 
     def fetch_deployment_availabilities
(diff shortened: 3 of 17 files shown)


abonas:
@miq-bot add_label events

app/models/manageiq/providers/hawkular/inventory/availability_updates.rb
```diff
@@ -0,0 +1,17 @@
...
+    def initialize(targets)
+      @targets = targets
+    end
+
+    def name
+      "Collection of availabilities to update on inventory entities"
+    end
+
+    def id
+      "Collection: #{@targets.map(&:manager_ref)}"
```
abonas: should the id be such a long string?

app/models/manageiq/providers/hawkular/inventory/availability_updates.rb
```diff
@@ -0,0 +1,17 @@
+module ManageIQ::Providers
+  class Hawkular::Inventory::AvailabilityUpdates
+    delegate :select, :to => :@targets
+
+    def initialize(targets)
+      @targets = targets
+    end
+
+    def name
+      "Collection of availabilities to update on inventory entities"
```
abonas: is the name displayed somewhere? if yes, it needs translation

app/models/manageiq/providers/hawkular/inventory/parser/availability_updates.rb
```diff
@@ -0,0 +1,22 @@
+module ManageIQ::Providers
+  class Hawkular::Inventory::Parser::AvailabilityUpdates < ManagerRefresh::Inventory::Parser
+    def parse
+      fetch_server_availabilities
+      fetch_deployment_availabilities
+    end
+
+    def fetch_server_availabilities
```
abonas: it looks like there's an opportunity here for code reuse.
because this method and the one below look exactly the same

app/models/manageiq/providers/hawkular/inventory/availability_updates.rb
israel-hdez: It doesn't need to be long. But, as `name`, it's only used when [writing logs here](https://github.com/ManageIQ/manageiq/blob/5edac5566820d4d4b14bf55bdcc75b1daca693e5/app/models/ems_refresh/refreshers/ems_refresher_mixin.rb#L23). I found no other usage of it. I'm trying to be verbose for logging.

app/models/manageiq/providers/hawkular/inventory/parser/availability_updates.rb
israel-hdez: Hmmm, I'll try to think a way to make it reusable.

israel-hdez: I couldn't think a way to refactor it :(

israel-hdez:
I think this is ready. Only reviews left. So, I will do:
@miq-bot remove_label wip

abonas:
@Jiri-Kremser please review. thanks

app/models/manageiq/providers/hawkular/middleware_manager/event_catcher/stream.rb
```diff
@@ -21,20 +23,113 @@ def each_batch
...
+
+    server_avails = fetch_server_availabilities(parser)
+    deploy_avails = fetch_deployment_availabilities(parser)
+
+    server_avails.concat(deploy_avails)
+  end
+
+  def fetch_server_availabilities(parser)
+    # For servers, it's also needed to refresh server state from inventory.
+    $mw_log.debug("#{log_prefix} Retrieving server state from Hawkular inventory")
```
Jiri-Kremser: nit: perhaps, `s/state/states/g` would be better

app/models/manageiq/providers/hawkular/inventory/collector/availability_updates.rb
```diff
@@ -0,0 +1,11 @@
+module ManageIQ::Providers
+  class Hawkular::Inventory::Collector::AvailabilityUpdates < ManagerRefresh::Inventory::Collector
+    def deployment_updates
+      @target.select { |item| item.association == :middleware_deployments }
+    end
+
+    def server_updates
+      @target.select { |item| item.association == :middleware_servers }
```
Jiri-Kremser: did you test your code for the WF running in domain mode?

israel-hdez: Yes, I tested domain mode. There is no need to make any distinction.

app/models/manageiq/providers/hawkular/middleware_manager/event_catcher/stream.rb
israel-hdez: It is now changed :)

abonas:
@Jiri-Kremser there's no milestone set for this PR

PR #29, opened 2017-07-19
run provider generator to sync changes from core

durandom:
ran the generator again with

```bash
bundle exec rails g provider <name> --path ../
```

these are the changes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,7 +9,6 @@ begin
 rescue LoadError
 end
 
-require 'manageiq-providers-hawkular'
 require 'bundler/gem_tasks'
 
 FileList['lib/tasks_private/**/*.rake'].each { |r| load r }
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -9,7 +9,6 @@ begin
 rescue LoadError
 end
 
-require 'manageiq-providers-hawkular'
 require 'bundler/gem_tasks'
```
abonas: why is it ok to replace this dependency?

abonas:
@durandom can you please explain what is this generator and why it should be rerun?

durandom:
@abonas the generator from core is used to scaffold a new provider. From time to time there are  some refactorings and just running it again is an easy way to keep the providers consistent.

Rakefile
durandom: because this is name of the gem and since its in the gemspec its autorequired

Fryguy:
> because it has a strong assumption that the provider will be a gem available in the rubygems.org. Btw. currently it creates a badge with a broken link, if the gem is not available there.

This is reasonable. Can you open this as an issue in manageiq where the generator lives?

PR #30, opened 2017-07-25
Adding JDR report model class with logic to generate reports

israel-hdez:
JDR reports are subordinated to middleware servers. To generate a JDR
report, two steps are required:

1. Enqueue a report to be generated. A helper method in MiddlewareServer
model class is included in order to enqueue the report.
2. Generate the JDR report. This is accomplished invoking
generate_jdr_report method. This method shouldn't be invoked directly. A
MiQ worker should call it automatically once the report is enqueued.

Binary data is stored into vmdb using BinaryBlob model.

JIRA: https://issues.jboss.org/browse/HAWKULAR-1231

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
@@ -1,8 +1,8 @@
 require 'timeout'
 
 module ManageIQ::Providers
-  class Hawkular::MiddlewareManager::MiddlewareJdrReport < ApplicationRecord
-    self.table_name = 'middleware_jdr_reports'
+  class Hawkular::MiddlewareManager::MiddlewareDiagnosticReport < ApplicationRecord
+    self.table_name = 'middleware_diagnostic_reports'
 
     STATUS_QUEUED = 'Queued'.freeze
     STATUS_RUNNING = 'Running'.freeze
@@ -40,7 +40,7 @@ module ManageIQ::Providers
       ready? || erred?
     end
 
-    def generate_jdr_report
+    def generate_diagnostic_report
       $mw_log.debug("#{log_prefix} Sending to Hawkular a request to generate JDR report [#{id}].")
 
       self.status = STATUS_RUNNING
@@ -84,7 +84,7 @@ module ManageIQ::Providers
         :class_name  => self.class.name,
         :instance_id => id,
         :role        => 'ems_operations',
-        :method_name => 'generate_jdr_report',
+        :method_name => 'generate_diagnostic_report',
         :msg_timeout => ::Settings.ems.ems_hawkular.jdr.generation_timeout.to_i_with_method + 30.seconds
       )
 
@@ -121,7 +121,7 @@ module ManageIQ::Providers
     end
 
     def log_prefix
-      @_log_prefix ||= "EMS_#{ems_id}(Hawkular::MWM::MwJdrReport)"
+      @_log_prefix ||= "EMS_#{ems_id}(Hawkular::MWM::MwDiagnosticReport)"
     end
   end
 end
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
@@ -88,6 +88,18 @@ module ManageIQ::Providers
         :msg_timeout => ::Settings.ems.ems_hawkular.jdr.generation_timeout.to_i_with_method + 30.seconds
       )
 
+      EmsEvent.add_queue(
+        'add', ems_id,
+        :ems_id          => ems_id,
+        :source          => 'EVM',
+        :timestamp       => Time.zone.now,
+        :event_type      => 'hawkular_event',
+        :message         => _('Generation of JDR report was requested by a user.'),
+        :middleware_ref  => middleware_server.ems_ref,
+        :middleware_type => 'MiddlewareServer',
+        :username        => requesting_user
+      )
+
       $mw_log.info("#{log_prefix} JDR report [#{id}] for server [#{middleware_server.ems_ref}] enqueued with job #{job.id}.")
     end
 
@@ -108,7 +120,9 @@ module ManageIQ::Providers
       end
 
       $mw_log.info("#{log_prefix} Generation of JDR report [#{id}] [#{binary_blob.name}] succeded.")
-      @finish_signal << STATUS_READY
+
+      # Unblock main thread at generate_diagnostic_report method
+      @finish_signal << :success
     end
 
     def jdr_report_failed(error)
@@ -117,7 +131,8 @@ module ManageIQ::Providers
       self.error_message = error
       save!
 
-      @finish_signal << STATUS_READY
+      # Unblock main thread at generate_diagnostic_report method
+      @finish_signal << :failure
     end
 
     def log_prefix
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_jdr_report.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_jdr_report.rb
@@ -1,0 +1,112 @@
+module ManageIQ::Providers
+  class Hawkular::MiddlewareManager::MiddlewareJdrReport < ApplicationRecord
+    self.table_name = 'middleware_jdr_reports'
+
+    STATUS_QUEUED = 'Queued'.freeze
+    STATUS_RUNNING = 'Running'.freeze
+    STATUS_ERROR = 'Error'.freeze
+    STATUS_READY = 'Ready'.freeze
+
+    belongs_to :middleware_server
+    has_one :binary_blob, :as => :resource, :dependent => :destroy
+
+    validates :middleware_server_id, :requesting_user, :presence => true
+    validates :status, :inclusion => { :in => [STATUS_QUEUED, STATUS_RUNNING, STATUS_ERROR, STATUS_READY] }
+
+    delegate :ext_management_system, :to => :middleware_server
+    delegate :ems_id, :to => :middleware_server
+    after_create :enqueue_job
+    before_create :set_queued_date
+
+    after_initialize do |item|
+      item.status = STATUS_QUEUED if new_record?
+    end
+
+    def queued?
+      status == STATUS_QUEUED
+    end
+
+    def ready?
+      status == STATUS_READY
+    end
+
+    def erred?
+      status == STATUS_ERROR
+    end
+
+    def ran?
+      ready? || erred?
+    end
+
+    def generate_jdr_report
+      $mw_log.debug("#{log_prefix} Sending to Hawkular a request to generate JDR report [#{id}].")
+
+      self.status = STATUS_RUNNING
+      save!
+
+      callback = proc do |on|
+        on.success(&method(:jdr_report_succeded))
+        on.failure(&method(:jdr_report_failed))
+      end
+
+      @connection = ext_management_system.connect.operations(true)
+      @connection.export_jdr(middleware_server.ems_ref, true, &callback)
+    end
+
+    private
+
+    def set_queued_date
+      self.queued_at = Time.current if queued?
+    end
+
+    def enqueue_job
+      unless queued?
+        $mw_log.debug("#{log_prefix} JDR report registry [#{id}] for server [#{middleware_server.ems_ref}] created with status [#{status}].")
+        return
+      end
+
+      job = MiqQueue.submit_job(
+        :class_name  => self.class.name,
+        :instance_id => id,
+        :role        => 'ems_operations',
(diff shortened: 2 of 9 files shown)


abonas:
@israel-hdez does this also require a schema change for a new table to be added? if yes, is there a PR linked to this with the db migration script?
Also, shouldn't this report be related/inherit the MiqReport ?

israel-hdez:
@abonas Yes, it requires a schema change. It's here: ManageIQ/manageiq-schema#45
Not sure if it should be related to MiqReport model. What I understand is that MiqReport is for cloud intelligence reports and if I add a record there, a new item may appear in cloud intelligence. I'll asks around to confirm.

israel-hdez:
@jrafanie Currently, which ruby version are we targeting? As you can see, I'm getting a syntax error for ruby 2.2. I can fix it with `@connection.close_connection! if @connection`, but I'll get a codeclimate warning suggesting to use safe navigation. Which warning/error should I ignore?

jrafanie:
> @jrafanie Currently, which ruby version are we targeting? As you can see, I'm getting a syntax error for ruby 2.2. I can fix it with @connection.close_connection! if @connection, but I'll get a codeclimate warning suggesting to use safe navigation. Which warning/error should I ignore?

We'll soon be dropping ruby 2.2 and only supporting 2.3 and 2.4 but we're not there yet.  Please use your suggestion for now.

abonas:
@miq-bot add_label enhancement, providers/middleware

abonas:
@miq-bot add_label providers/hawkular

israel-hdez:
@abonas Review round can be started.

mtho11:
@israel-hdez can we get CI to pass?

israel-hdez:
@mtho11 Now CI is passing.
@miq-bot remove_label wip

mtho11:
@israel-hdez great! Code climate issues is a false positive give we can't use the '&.' operator yet for ruby 2.2.x support as you well know.

mtho11:
@israel-hdez great! Code climate issues is a false positive give we can't use the '&.' operator yet for ruby 2.2.x support as you well know.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -0,0 +1,139 @@
+require 'timeout'
+
+module ManageIQ::Providers
+  class Hawkular::MiddlewareManager::MiddlewareDiagnosticReport < ApplicationRecord
+    self.table_name = 'middleware_diagnostic_reports'
+
+    STATUS_QUEUED = 'Queued'.freeze
+    STATUS_RUNNING = 'Running'.freeze
+    STATUS_ERROR = 'Error'.freeze
+    STATUS_READY = 'Ready'.freeze
```
josejulio: We could use [rails enums](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) to simplify the logic.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -0,0 +1,139 @@
...
+
+    belongs_to :middleware_server
+    has_one :binary_blob, :as => :resource, :dependent => :destroy
+
+    validates :middleware_server_id, :requesting_user, :presence => true
+    validates :status, :inclusion => { :in => [STATUS_QUEUED, STATUS_RUNNING, STATUS_ERROR, STATUS_READY] }
```
josejulio: We could use [rails enums](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) to simplify the logic.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -0,0 +1,139 @@
...
+    delegate :ext_management_system, :to => :middleware_server
+    delegate :ems_id, :to => :middleware_server
+    after_create :enqueue_job
+    before_create :set_queued_date
+
+    after_initialize do |item|
+      item.status = STATUS_QUEUED if new_record?
+    end
+
+    def queued?
```
josejulio: We could use [rails enums](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) to simplify the logic.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -0,0 +1,139 @@
...
+      status == STATUS_QUEUED
+    end
+
+    def ready?
```
josejulio: We could use [rails enums](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) to simplify the logic.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -0,0 +1,139 @@
...
+      status == STATUS_READY
+    end
+
+    def erred?
```
josejulio: We could use [rails enums](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) to simplify the logic.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -0,0 +1,139 @@
...
+      status == STATUS_ERROR
+    end
+
+    def ran?
```
josejulio: We could use [rails enums](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) to simplify the logic.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -0,0 +1,139 @@
...
+      @finish_signal << STATUS_READY
+    end
+
+    def jdr_report_failed(error)
+      $mw_log.warn("#{log_prefix} Generation of JDR report [#{id}] failed: #{error}.")
+      self.status = STATUS_ERROR
+      self.error_message = error
+      save!
+
+      @finish_signal << STATUS_READY
```
josejulio: What is this queue for, and why are we queuing STATUS_READY on fail?

mtho11:
Thanks for the review @josejulio I know you have worked on these reports before ;)

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
israel-hdez: The queue `@finish_signal` is for blocking the main thread when the request to generate JDR is sent to Hawkular-services command gateway [right here](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/30/files#diff-c6947bd67b4592e6ae39ce22e286f625R58). You know: because websockets are being used for cmd-gw operations, `export_jdr` method is asynchronous. The callbacks are ran in the context of a secondary thread and in order to unblock the main thread I just pushed something to `@finish_signal`. It could have been any arbitrary value, it doesn't mean anyting nor it's used. The only purpose is to unblock the main thread once JDR fails or succeeds to close the websocket connection [here](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/30/files#diff-c6947bd67b4592e6ae39ce22e286f625R68) and ensure that the secondary thread is terminated.

Should the value cause confusion, I could replace it with an arbitrary `string`, a `:symbol` or a boolean value.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
israel-hdez: Hmm... I have my doubts. Rails enums are mapped to integers in the database and most of the models I have seen, are using strings if they have statuses associated to them (probably to be verbose when querying directly the database). See, for example, [MiqQueue](https://github.com/ManageIQ/manageiq/blob/ef367b7e2e34750d3142494c42327c5c13da3dfb/app/models/miq_queue.rb#L75), [MiqTask](https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_task.rb#L3).

Although, if those models were written before rails supported enums, that could explain why they are coded like that.

Hmm... Schema PR is already merged, so the column is a string. I could try to see if enums work properly with a database string column to avoid having to place another PR to "fix" the column.

israel-hdez: This also would lead some changes to UI PR, because is't displaying the status as is: https://github.com/ManageIQ/manageiq-ui-classic/pull/1764/files#diff-002598c1d95530eae2e85c719166439fR63

josejulio: I suppose is OK if is not possible with rails enums.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
josejulio: Oh ok, it wasn't clear to me.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
israel-hdez: It doesn't work with the string database field :(

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
israel-hdez: Changed and placed a comment to be more clear.

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
josejulio: Well, we could either leave as it or change to use rails enums. It's your choice now :)
We might as well implement (or use) something on the future that supports string enums.
@mtho11 @abonas WDYT?

mtho11: I'm a fan of enums!

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -108,7 +120,9 @@ def jdr_report_succeded(data)
       end
 
       $mw_log.info("#{log_prefix} Generation of JDR report [#{id}] [#{binary_blob.name}] succeded.")
-      @finish_signal << STATUS_READY
+
+      # Unblock main thread at generate_diagnostic_report method
+      @finish_signal << 'https://youtu.be/-Elk_9fVo9E'
```
mtho11: What are these youtube links?

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -117,7 +131,8 @@ def jdr_report_failed(error)
       self.error_message = error
       save!
 
-      @finish_signal << STATUS_READY
+      # Unblock main thread at generate_diagnostic_report method
+      @finish_signal << 'https://youtu.be/jkRN9zcLH1s'
```
mtho11: And this youtube link?

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
israel-hdez: Oh, sorry. I was in a moment doing dumb things. Fixed.

israel-hdez: Also, see https://github.com/ManageIQ/manageiq-providers-hawkular/pull/30#discussion_r135082037

josejulio: Is an easter :egg:

josejulio: It was cool, sadly is gone :(

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
abonas: @israel-hdez you mentioned blocking the main thread above - I did not follow all the discussion, but is this blocking the entire miq?

israel-hdez: @abonas No. By "main thread" I mean the miq worker that is running the operation, which will wait until JDR operation completes (or timeouts). All other miq workers will keep running normally.

PR #31, opened 2017-07-26
Adding link to the vm also for the servers running in the domain mode

Jiri-Kremser:
https://bugzilla.redhat.com/show_bug.cgi?id=1404270

note: I'd make the `find_host_by_bios_uuid` method private, but it's also being used from specs


app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -44,10 +44,7 @@ def fetch_middleware_servers
                 set_lives_on(server, container) if container
               end
             else
-              machine_id = collector.machine_id(eap.feed)
-              host_instance = find_host_by_bios_uuid(machine_id) ||
-                              find_host_by_bios_uuid(alternate_machine_id(machine_id))
-              set_lives_on(server, host_instance) if host_instance
+              associate_with_rhevm(server, eap.feed)
```
abonas: is this linking only to VMs in rhevm? afaik the functionality of linkage existing prior to this patch works for vmware VMs as well.

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -95,6 +92,8 @@ def fetch_domain_servers(feed)
           server = persister.middleware_servers.find_or_build(domain_server.path)
           parse_middleware_server(domain_server, server, true, server_name)
 
+          associate_with_rhevm(server, feed)
```
abonas: I migt be missing here something, but is this matching using the same feed for all the servers?

abonas:
do you plan to add tests for this new functionality?

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
Jiri-Kremser: Per each feed, we list the domain servers and then for this `<server, feed>` pair we try add the association by running this method. The `associate_with_rhevm` figures out the `machine_id` (it was stored under the operating system resource in the inventory, if I recall correctly) and based on this id the linking is done. This will work for domain mode that has only one host - the default scenario. For multiple hosts, it will show that the servers are running on the domain controller host, but this wont be the true. Unfortunately, we don't have this info (machine id per host controller) in the inventory, so this can be improved in the future.

abonas:
@Jiri-Kremser please update PR title not to be RHEV specific
@israel-hdez @rubenvp8510 please re-review the latest revision

Jiri-Kremser:
> do you plan to add tests for this new functionality?

I've added specs

vprusa:
@abonas you can find polarion test cases listed at [this trello card](https://trello.com/c/oBeMkciI/254-inv-008-view-relationships-between-eap-servers-and-vms-that-run-a-linux-os-and-navigate-across-the-relationships), automation is now pending on setting up rhevm environment

abonas:
@israel-hdez @rubenvp8510 can you please re-review? if it's ready I can merge it

israel-hdez:
@abonas Code looks good. I think it can be merged.

Issue #35, opened 2017-08-28
Cross Link to VM - Not Shown


hhovsepy:
In Details page of EAP7 Standalone/Domain mode servers, cross link to VMs are not shown.

1. Add new Infrastructure Provider of type RHEVM. Enter "admin@internal" user's credentials.
Make sure that VMs have got IP Address and Hostnames.
2. On a VM install EAP7 in Standalone or in Domain mode, configure hawkular agent to Hawkular Services.
3. Add Hawkular Services as Middleware Provider.
4. Go to Middlewar Server Details page.
Cross link to VM is not there.

Attached screenshots for VMs and EAP7 Servers details pages for Domain and Standalone modes.

![vm_eap7_domain](https://user-images.githubusercontent.com/604313/29761747-126d8560-8bcb-11e7-9f54-052c06d86cc4.png)
![vm_eap7_standalone](https://user-images.githubusercontent.com/604313/29761746-126bbbd6-8bcb-11e7-82bd-cc28322c6ec7.png)
![eap7_standalone](https://user-images.githubusercontent.com/604313/29761745-126b1456-8bcb-11e7-8e41-db2759387afd.png)
![eap7_domain](https://user-images.githubusercontent.com/604313/29761744-126935d2-8bcb-11e7-986c-394500db228a.png)

hhovsepy:
tested on version master.20170828012921_

hhovsepy:
evm.log
[vm_evm.zip](https://github.com/ManageIQ/manageiq-providers-hawkular/files/1255987/vm_evm.zip)

abonas:
the log is huge, so without focusing on specific timeframe it's very difficult to find there something.
lets try something else instead first:
please go to miq db, and do a select on table vms (where name is the name of the vm you expect to link with)
also, please select the row that represents the eap server from middleware_servers table

hhovsepy:
attaching vms and middleeware_servers tables all columns for EAP7 standalone mode server:

[db_vms.txt](https://github.com/ManageIQ/manageiq-providers-hawkular/files/1256652/db_vms.txt)
[db_servers.txt](https://github.com/ManageIQ/manageiq-providers-hawkular/files/1256653/db_servers.txt)

abonas:
@israel-hdez can you please have a look.
[...]

PR #37, opened 2017-08-29
Fixes VMs and MW servers not being cross-linked when GUIDs differ only because of dashes

israel-hdez:
Fixes issue #35. VMs and MW servers are not being cross-linked because
Hawkular Agent is reporting Machine ID as a 32 character string - a GUID
withouth dashes -, while RHVM provider is reporting machine ids as 36
character strings - GUID with dashes in standard locations.

This patch is adding dashes at standard locations to machine id reported
by Hawkular Agent, if needed.

Also, collector should not escape feed id when retrieving OS resource
types. This was preventing cross-linking to happen for MW servers in
domain mode.

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
@@ -122,6 +122,23 @@ module ManageIQ::Providers
         alternate.join('-')
       end
 
+      # Add standard dashes to a machine GUID that doesn't have dashes.
+      #
+      # @param machine_id [String] the GUI to which dashes should be added. The string
+      #   is validated to be 32 characters long and to only contain valid hexadecimal
+      #   digits. If any validations fail, nil is returned.
+      # @return [String] the GUI with dashed added at standard locations.
+      def dashed_machine_id(machine_id)
+        return nil if machine_id.nil? || machine_id.length != 32 || machine_id[/\H/]
+        [
+          machine_id[0, 8],
+          machine_id[8, 4],
+          machine_id[12, 4],
+          machine_id[16, 4],
+          machine_id[20, 12]
+        ].join('-')
+      end
+
       def swap_part(part)
         # here, we receive parts of an UUID, split into an array with 2 chars each element, and reverse the invidual
         # elements, joining and reversing the final outcome
@@ -327,7 +344,8 @@ module ManageIQ::Providers
         # Add the association to vm instance if there is any
         machine_id = collector.machine_id(feed)
         host_instance = find_host_by_bios_uuid(machine_id) ||
-                        find_host_by_bios_uuid(alternate_machine_id(machine_id))
+                        find_host_by_bios_uuid(alternate_machine_id(machine_id)) ||
+                        find_host_by_bios_uuid(dashed_machine_id(machine_id))
         set_lives_on(server, host_instance) if host_instance
       end
 
--- a/spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
+++ b/spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
@@ -279,6 +279,20 @@ describe ManageIQ::Providers::Hawkular::Inventory::Parser::MiddlewareManager do
     end
   end
 
+  describe '#dashed_machine_id' do
+    it 'should reject id that is not 32 charasters length' do
+      expect(parser.dashed_machine_id('abc123')).to be_nil
+    end
+
+    it 'should reject id with non hexadecimal characters' do
+      expect(parser.dashed_machine_id('abcdef1234567890abcdef123456789P')).to be_nil
+    end
+
+    it 'should add dashes at standard locations' do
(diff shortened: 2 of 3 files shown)


israel-hdez:
@miq-bot add_label bug, inventory

abonas:
@israel-hdez 
what change caused this regression? on agent side?

israel-hdez:
@abonas Not sure, so I cannot say if a change broke the cross-linking. However, IMO it's not exactly a regression. To the end user, the issue is the same but the underlying cause (apparently) isn't the same.

I saw in the code a link to a previous issue: https://bugzilla.redhat.com/show_bug.cgi?id=1294461 but I don't have access to it. However, the [code that solves it](https://github.com/israel-hdez/manageiq-providers-hawkular/blob/0fca6e947d5db8ac982d542f27b22faf5bcc519c/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L107) is dealing with something different regarding how machine-ids are handled. This is because i think is not exactly a regression.

The agent may not be the problem because is reporting machine-ids "as-is" by reading system files or env variables. But there are other parts like the OS of the VM, the virtualization engine and the miq vm-provider. All parts must report machine-ids consistently to have the cross-linking working correctly, but that's not happening always.

So, in the provider we are trying to find the underlying vm by testing all know variants of the same machine id. With this fix, we now have three variants: https://github.com/ManageIQ/manageiq-providers-hawkular/pull/37/files#diff-a6822e8aa662958bb0ab11a0ec7cc558R346

abonas:
> I saw in the code a link to a previous issue: https://bugzilla.redhat.com/show_bug.cgi?id=1294461 but I don't have access to it.

have you logged in? no access even after authenticating?

israel-hdez:
@abonas No access even after authenticating.

israel-hdez:
@abonas Now I can confirm that the previous code is dealing with a different issue. The previous code is only a workaround while the bugzilla was fixed. It's dealing with an endianess problem and, also, is adding dashes to the `machine_id`. But now that the endianess bug is fixed in the Linux Kernel, only dashes are needed and that's what is doing this fix.

The endianess fix was released in Linux Kernel 4.4, according to https://bugzilla.kernel.org/show_bug.cgi?id=109731#c3 and propagated later to RHEL.

israel-hdez:
would you like to review/merge @abonas @mtho11?

abonas:
@cfcosta please review

abonas:
@xeviknal I know you are not familiar with this yet to be a reviewer, but I wanted you to have a look and get a bit familiar with the cross linking feature via this PR as well as with our tests. if you need any clarifications on the feature itself, @israel-hdez can elaborate a bit :)

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -122,6 +122,23 @@ def alternate_machine_id(machine_id)
...
 
+      # Add standard dashes to a machine GUID that doesn't have dashes.
+      #
+      # @param machine_id [String] the GUI to which dashes should be added. The string
+      #   is validated to be 32 characters long and to only contain valid hexadecimal
+      #   digits. If any validations fail, nil is returned.
+      # @return [String] the GUI with dashed added at standard locations.
+      def dashed_machine_id(machine_id)
+        return nil if machine_id.nil? || machine_id.length != 32 || machine_id[/\H/]
+        alternate = []
```
cfcosta:
```ruby
[machine_id[0,8], machine_id[8,4], machine_id[12,4]. machine_id[16,4], machine_id[20,12]].join('-')
```

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
```diff
@@ -279,6 +279,20 @@ def call_subject(feeds = %w(f1), resources = [stubbed_resource])
     end
   end
 
+  describe 'dashed_machine_id' do
```
cfcosta: This should be `describe '#dashed_machine_id'`, for consistency.

spec/models/manageiq/providers/hawkular/inventory/parser/middleware_manager_spec.rb
```diff
@@ -279,6 +279,20 @@ def call_subject(feeds = %w(f1), resources = [stubbed_resource])
...
+    it 'should reject id that is not 32 charasters length' do
+      expect(parser.dashed_machine_id('abc123')).to be_nil
+    end
+
+    it 'should reject id with non hexadecimal characters' do
+      expect(parser.dashed_machine_id('abcdef1234567890abcdef123456789P')).to be_nil
+    end
+
+    it 'should add dashes at standard locations' do
+      expect(parser.dashed_machine_id('abcdef1234567890abcdef1234567890')).to be == 'abcdef12-3456-7890-abcd-ef1234567890'
```
cfcosta: `expect(foo).to eq bar`, the error messages are better that way.

abonas:
@israel-hdez @cfcosta let me know when it's ready for merge

israel-hdez:
You have my :+1: 
@cfcosta any other thoughts?
@xeviknal did you give it a look?

cfcosta:
@israel-hdez @abonas LGTM

PR #38, opened 2017-09-04
Add setup environment guide to README

cfcosta:
[skip ci]
This should simplify the lifes of anyone wanting to contribute.


cfcosta:
@abonas regarding your comments on the gist:

>why running bin/update is necessary on all repos ? not only on the core repo?

Right now it updates the `spec/manageiq` clone of the repo, so it's necessary. I think it might be worth adding a section on the guide to also guide on how to symlink the manageiq used on tests to the same being used in development, what do you think?

>adding a link to the manageiq developer guide is misleading as it describes a different method of work with the other repos (via the plugins folder) - it can be confusing for newcomers that might attempt to do both ways

That's true, should I copy the environment setup part to our side as well?

abonas:
> That's true, should I copy the environment setup part to our side as well?

that will require us to maintain this doc in addition to the main guide which can be a burden

README.md
```diff
@@ -14,9 +14,348 @@ ManageIQ plugin for the Hawkular provider.
...
+
+It will ask then some questions:
+
+* What should it use for running the services? Docker-Compose? Openshift?
+* What version of Hawkular?
+* What version of Cassandra?
+* What WildFly version?
+* Standalone or Domain mode for Wildfly?
+* How many WildFly servers?
+
```
abonas: this is great. I think currently missing:
1. information on configuring mutable vs immutable servers - as in miq we have functionality specific to each case
2. how to start an additional server on an already running env

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+For OS X, `rbenv` and `ruby-build` are available on homebrew, and can be installed like this:
+
+```bash
+brew upgrade rbenv ruby-build
+echo 'eval "$(rbenv init -)"' >> ~/.bashrc
+eval "$(rbenv init -)"
+```
+
+For Linux, it will have to be installed via default git checkout:
```
abonas: isn't it easier with rvm?

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+
+### Updating the repositories
+
+To update the repositories, you need to go on each one and run `bin/update`. This script takes care of it:
+
+```bash
+#!/bin/bash
+
+MIQ_DIR="$HOME/manageiq"
+OLD_PWD="$(pwd)"
```
abonas: what is this for?

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+git remote add upstream git@github.com:ManageIQ/manageiq-ui-classic
+cd ..
+
+cd manageiq-schema
+git remote add upstream git@github.com:ManageIQ/manageiq-schema
+cd ..
+
+cd manageiq-api
+git remote add upstream git@github.com:ManageIQ/manageiq-api
+cd ..
```
abonas: can't all cd point to the next folder already without first going to ".." ?

e.g. 
 +cd manageiq-api
 +git remote add upstream git@github.com:ManageIQ/manageiq-api
 +cd ../manageiq-providers-hawkular

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+Any Linux/OS X System is compatible, as far as we are concerned.
+
+### Installing Dependencies
+
+#### Fedora
+
+First, install all system dependencies:
+
+```bash
+sudo dnf -y git-all memcached postgresql-devel postgresql-server postgresql-contrib bzip2 libffi-devel readline-devel sqlite-devel nodejs gcc-c++ libcurl-devel npm openssl-devel cmake openscap
```
abonas: docker missing here for install?

josejulio: Isn't this all (or part of) already here? https://github.com/ManageIQ/guides/blob/master/developer_setup.md

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+
+There are two ways of installing Hawkular: manually or via Docker. There are
+use cases for both, and you can mix and match for your needs, but the simplest
+way of getting it working is via Hawkinit, which downloads, configures and runs
+the hawkular server, and 0+ wildfly instances with the Hawkular agent installed
+and configured.
+
+#### Via Docker (recommended)
+
+First, we need to install [hawkinit](https://github.com/Jiri-Kremser/hawkinit),
```
abonas: isn't this on https://github.com/hawkular/hawkinit ?

josejulio: Yes, we should link there and provide hawkinit recipes for common scenarios.

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+`docker` group:
+
+```bash
+sudo usermod -a -G docker `whoami`
+newgrp docker # to update your groups on the session
+```
+
+Then, by running `hawkinit`, you'll see something like this:
+
+![](screenshots/hawkinit.png)
```
abonas: screenshot needs fixing after [this](https://github.com/hawkular/hawkinit/pull/38) will be merged

README.md
abonas: yes, I commented about it [here](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/38#issuecomment-327071406)

README.md
josejulio: I'm not sure, I think is matter of choice. I use `chruby`. 
If rvm setup is easier we should provide that one here instead, unless there is a real advantage of using rbenv.

README.md
josejulio: I added some info here: https://github.com/cfcosta/manageiq-providers-hawkular/pull/1

README.md
abonas: @cfcosta ^^

README.md
abonas: @cfcosta ^^

README.md
abonas: @josejulio but docker is missing in that dev guide because it is relevant for hawkinit but not for miq at the moment

README.md
abonas: @cfcosta [this](https://github.com/hawkular/hawkinit/pull/38) has been merged, please update the screenshot

abonas:
please add [skip ci] in the top comment

README.md
abonas: maybe it's still worth to have a guide of our own because we can update it as frequent as we want and make it tailored for our needs

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
 
 ## Development
 
-See the section on pluggable providers in the [ManageIQ Developer Setup](http://manageiq.org/docs/guides/developer_setup)
+This is a quick guide outlining all the necessary steps to get ManageIQ up and
```
israel-hdez: I have the idea that this doesn't fit well in the `README.md`. This is converting the readme file into a developer quick start, because readme doesn't have a lot of info. I think it's better to have this guide in its own file and place a link here.

README.md
israel-hdez: Instead of cd-ing, you could also pass -C flag to simplify:

`git -C manageiq-ui-classic/  remote add upstream git@github.com:ManageIQ/manageiq-ui-classic`

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+```ruby
+gem 'pry'
+gem 'pry-byebug'
+```
+
+The next step is to configure each repository:
+
+```bash
+cd ~/ManageIQ/manageiq
+./bin/setup
```
israel-hdez: isn't enough to run only this `bin/setup`? The other ones aren't needed unless you are running tests. But in order to run tests, some extra setup is required.

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+
+cd hawkular-client-ruby
+bundle install
+cd ..
+```
+
+This should be enough. To run the application, we can run this, which will run the application on background on port 3000:
+
+```bash
+bundle exec rake evm:start
```
israel-hdez: You should be in `manageiq` directory (main repo).

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+Sign in with the default credentials (`admin`/`smartvm`), and you should be good to go!
+
+### Updating the repositories
+
+To update the repositories, you need to go on each one and run `bin/update`. This script takes care of it:
+
+```bash
+#!/bin/bash
+
+MIQ_DIR="$HOME/manageiq"
```
israel-hdez: This should be `MIQ_DIR="$HOME/ManageIQ"` to be consistent with directory name at [line 138](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/38/files#diff-04c6e90faac2675aa89e2176d2eec7d8R138).

README.md
```diff
@@ -14,9 +14,345 @@ ManageIQ plugin for the Hawkular provider.
...
+    git fetch -a
+    git pull --rebase upstream master
+    ./bin/update
+    cd -
+  fi
+done
+
+echo "Updating core..."
+
+cd $MIQ_DIR/core
```
israel-hdez: This should be `cd $MIQ_DIR/manageiq` to be consistent with name used at [line 127](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/38/files#diff-04c6e90faac2675aa89e2176d2eec7d8R127)

cfcosta:
@abonas updated.

README.md
abonas: sure, we can move this to another file.

abonas:
@cfcosta are you sure you pushed the changes? I don't see a lot of questions/comments addressed.
for example: the missing docker or regarding rvm or moving this to another file or changing the screenshot to a newer one where typo was fixed

app/models/manageiq/providers/hawkular/middleware_manager/event_catcher/runner.rb
```diff
@@ -42,13 +42,13 @@ def monitor_events
       event_monitor_running
 
       # Separate alerts from avail updates
-      avail_updates, events = events.partition { |e| !e.kind_of?(::Hawkular::Alerts::Alert) }
+      events, avail_updates = events.partition { |e| !e.kind_of?(::Hawkular::Alerts::Event) }
```
josejulio: Is this file needed for the README.md in any way or was pushed by mistake?

josejulio:
Hawkinit changes on the Readme seems to be overwritten (git push -f?)
Can you please cherry pick https://github.com/cfcosta/manageiq-providers-hawkular/pull/1/commits/0e46f0fd2760198f97192cae7222838ebb1f4411 ?

README.md
cfcosta: There is. Afaik, if that didn't change, rvm overwrites system commands like cd and is generally a pain to work with, requiring implodes every now and then. I've been using rbenv since and the workflow couldn't be nicer.

Either way, I'm going to remove installation instructions and let everyone chooses what they want to use.

README.md
cfcosta: Also added a guide to Docker.

README.md
cfcosta: I just installed and the UI is the same, am I missing something?

README.md
cfcosta: I disagree, having ManageIQ working correctly is essential to test/developer on our provider code.

README.md
cfcosta: What kind of extra setup?

app/models/manageiq/providers/hawkular/middleware_manager/event_catcher/runner.rb
cfcosta: Pushed by mistake.

README.md
josejulio: Yes, see my commit. hawkinit is under hawkular organization now.

cfcosta:
@abonas rebased and added a Docker guide as well, can you check if there's anything else missing?

README.md
josejulio: I suppose they mean to have it more organized like this: https://github.com/ManageIQ/manageiq/blob/master/README.md

abonas:
it should be skip ci, not ci skip :)

abonas:
@cfcosta seems that @josejulio changes did not get in? please check
and also see the suggestion to link to this file from readme instead of it being *the* readme

README.md
israel-hdez: This should be: http://www.manageiq.org/docs/guides/developer_setup/plugins#running-tests

README.md
```diff
@@ -14,9 +14,336 @@ ManageIQ plugin for the Hawkular provider.
...
+cd $MIQ_DIR/manageiq
+
+./bin/update
+
+cd $OLD_PWD
+
+echo "DONE."
+```
+
+## Installing Docker
```
israel-hdez: I'm not very, very sure because I don't remember well... But IIRC, docker-compose is installed separately from the engine. That should be mentioned here because hawkinit depends on compose.

cfcosta: That's true, will add it.

mtho11:
This getting hard to follow with all the comments now. Let's merge and get a more contiguous version. But keep up with the commentary.

README.md
cfcosta: Afaik the `spec:setup` task is also run when you run `bin/setup`, so we should be covered there.

mtho11:
@abonas I think we should merge what we have and keep adding to it.

abonas:
merging. I see currently some work to be done from the comments above:
1. fix screenshot of hawkinit to a new one without typo
2. move this to another page and link from readme
3. @josejulio is your contribution to @cfcosta fork here or still ovewritten? (that's why I originally commented it's better to put it directly here to avoid overrides :))

josejulio:
> merging. I see currently some work to be done from the comments above:
> 
> fix screenshot of hawkinit to a new one without typo - this issue is tracking it
> @josejulio is your contribution to @cfcosta fork here or still ovewritten? (that's why I originally commented it's better to put it directly here to avoid overrides :))

Is here.

> @josejulio is there a way to document commands how to add a new additional eap server to the documented setup?

Sure, will add the info

PR #47, opened 2017-09-12
Fix mutable flag in hawkinit scripts

abonas:


abonas:
@josejulio @cfcosta @israel-hdez  please review

PR #48, opened 2017-09-14
Add autoload support to the engine

cfcosta:


cfcosta:
@abonas @israel-hdez can you help me test this? This should fix the problem of us having to restart the server after every change.

How to test:
- Start the server.
- Change something.
- See if that something changed correctly.

abonas:
@jrafanie @fryguy does it make sense to apply a similar solution in all split repos?

lib/manageiq/providers/hawkular/engine.rb
```diff
@@ -3,6 +3,11 @@ module Providers
     module Hawkular
       class Engine < ::Rails::Engine
         isolate_namespace ManageIQ::Providers::Hawkular
+
+        initializer('hawkular.autoload') do |app|
+          gem_root = File.expand_path('../../../..', __dir__)
+          config.autoload_paths += Dir["#{gem_root}/{app,lib}/**/"]
```
jrafanie: @durandom can this be put in the provider generator?  Do we want to autoload all directories under lib?  

Isn't app already autoloaded @cfcosta  for engines?

israel-hdez:
@cfcosta I'm not sure if I understand. If I make a change in the code, should the running app take with the new changes? Should this work in all contexts?... Workers doesn't seem to take the new changes. The UI for sure does.

lib/manageiq/providers/hawkular/engine.rb
cfcosta: Nope, not that I'm aware of. I'll do some more testing, because it seems that only UI changes are actually autoloaded, if any. Would be good to have that enabled for workers as well.

durandom: `ActiveSupport::Dependencies.autoload_paths` includes the `app/models` directory of every provider.

If you need `lib` or anything else, I suggest to add it like the UI does [here](https://github.com/ManageIQ/manageiq-ui-classic/blob/5fac2d1037453493187a6df0c4104e8b4fc0047f/lib/manageiq/ui/classic/engine.rb#L34)

cfcosta:
Closing because it should be able to reload workers, and this doesn't do it.

PR #49, opened 2017-09-14
Setup test environment for VCR recordings

israel-hdez:
This PR is meant to make it easier to setup an environment that can be used to run tests and record VCR cassettes.

It also includes some changes to enable live tests in CI.


israel-hdez:
Although I'm not sure about the live tests in CI. "Server broke connection" errors are common in Travis and so the need to restart travis jobs to make them pass.

lib/tasks_private/spec_hawkular.rake
```diff
@@ -0,0 +1,136 @@
+require 'net/http'
+require 'json'
+require 'hawkularclient'
+
+module ManageIQ::Providers::Hawkular::TaskHelpers
+  class LiveSetup
```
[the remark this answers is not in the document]
israel-hdez: Being honest, I also had my doubts about how to organize the code. At the end, I took the approach I saw at [evm_settings.rake](https://github.com/ManageIQ/manageiq/blob/master/lib/tasks/evm_settings.rake) where they have the module in the rake file. It make sense because that module is relevant only for the accompanying rake task.

But I think having a `task_helpers` directory seems reasonable. Let me know if you are better with that approach so that create the directory and I move the module there.

israel-hdez:
There is a Travis error that pops up too frequently: https://travis-ci.org/ManageIQ/manageiq-providers-hawkular/jobs/275600024#L2615

Probably we should skip the live tests in Travis and only use them in our local envs?

xeviknal:
I have been playing with this branch, and it makes really easy to do the integration tests.
From my point of view, it would be cool to have in the CI fully working. Do we have any master on Travis CI, here? It would be helpful.
Otherwise, I would go without the live spec in CI.

israel-hdez:
For now, I've reverted the `.travis.yml` file to leave aside the CI live tests and use them only in our envs when needed. Hopefully, we enable CI live tests in another PR, if it's somehow possible.

Looks like this PR is already approved, so I leave to you @mtho11 the decision to merge (or not).

josejulio:
> There is a Travis error that pops up too frequently: https://travis-ci.org/ManageIQ/manageiq-providers-hawkular/jobs/275600024#L2615
> 
> Probably we should skip the live tests in Travis and only use them in our local envs?

I think that `Server broke connection` is related to `websocket-client-simple` used in the hawkular-client
@rubenvp8510 might know more about this, as he was working on replacing it https://github.com/hawkular/hawkular-client-ruby/pull/205

ci_files/Dockerfile-wf-standalone
```diff
@@ -0,0 +1,8 @@
+FROM hawkular/wildfly-hawkular-agent:latest
```
josejulio: We need to move away from wildfly-hawkular-agent and use the javaagent instead.

rubenvp8510: > There is a Travis error that pops up too frequently: https://travis-ci.org/ManageIQ/manageiq-providers-hawkular/jobs/275600024#L2615
> 
> Probably we should skip the live tests in Travis and only use them in our local envs?
> I think that Server broke connection is related to websocket-client-simple used in the hawkular-client
> @rubenvp8510 might know more about this, as he was working on replacing it hawkular/hawkular-client-ruby#205

This not seems to be related to the WebSocket library, at least not directly, that library has a problem because it closes the socket before kills the polling thread and sometimes try to read from a closed socket (with getc) and that causes a segfault) this is not the case.

Here for some reason, the server is ending the connection or there are some communication issues before ending the request.

mtho11:
@israel-hdez if you resolve the conflicts it looks like this can be merged

ci_files/Dockerfile-wf-standalone
mtho11: @josejulio is this good to merge?

josejulio: wildfly-hawkular-agent won't be supported for inventory v4, we should use `hawkular/wildfly-hawkular-javaagent` The config is almost the same (yaml instead of xml).

Hawkular-services already uses the javaagent. (even in inventory v3)

israel-hdez:
@mtho11 @josejulio Solved the conflict. Also, now using javaagent docker image.

PR #53, opened 2017-09-20
Readme improvements

xeviknal:
Here my humble addition to this handy onboarding guide.
Most problems came up when using hawkinit. I did couple of pull request proposing fixes.
As well, here it is added a troubleshooting page for all those people who wants to share trouble they have. I started with the `npm install` @karelhala suggested last team meeting.

Feel free to discuss on everything.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -75,11 +75,11 @@ gem install bundler --no-ri --no-rdoc
 The first necessary step is to fork the following repositories by running this command:
 
 ```bash
-curl -u '<your_github_user>' https://api.github.com/repos/ManageIQ/manageiq/forks  
-https://api.github.com/repos/ManageIQ/manageiq-ui-classic/forks
-https://api.github.com/repos/ManageIQ/manageiq-schema/forks
-https://api.github.com/repos/ManageIQ/manageiq-api/forks
-https://api.github.com/repos/ManageIQ/manageiq-providers-hawkular/forks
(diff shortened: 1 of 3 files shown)


xeviknal:
@abonas here are the additions for the onboarding guides.
In general, in the PR, should I tag somebody to check it?

abonas:
> In general, in the PR, should I tag somebody to check it?

great. yes, typically we tag 1-2 relevant people on each PR, with time it's easier to know who is the relevant person :)
in our repo here in particular, all the maintainers also get a notification via mail on a new PR or issue

abonas:
@karelhala is the npm install troubleshooting temporary or permanent? e.g. is this going to go away in a month due to some specific change or expected to bump into all the time?

karelhala:
@abonas the problem is that it might happen or not, I honestly don't know why, but I blame rake tasks to be messed up. If you'd look on what `bin/update` does it runs `yarn` (which is just newer and fancier way of installing node packages), but for some reasons it won't install all packages and you end up with broken environment. So I suggest to keep it here, so we any newcomer won't get lost and knows what might happen.

abonas:
Ack @karelhala. Merging. Can u in addition perhaps involve core ui team to have a look?

xeviknal:
@karelhala couple of things:
1. could you please tell me where `yarn` is run on `bin/update`? Is in master branch?
2. In this doc file, it is not said to run `bin/update`, but `bin/setup` instead. Maybe this is why npm install it is not performed.

cc/ @abonas

PR #54, opened 2017-09-21
Update Getting Started doc with Arch Linux specific info

tumido:
Add a section about ManageIQ dependencies, how to install them, and how to configure PostgreSQL.
Add a section about Docker CE installation.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -49,6 +49,40 @@ sudo systemctl start postgresql
 sudo su postgres -c "psql -c \"CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'\""
 ```
 
+### Arch Linux
+
+First, install all system dependencies:
+```bash
+sudo pacman -S git memcached postgresql bzip2 libffi readline sqlite nodejs gcc
+# From AUR (i.e. via pacaur)
+pacaur -S openscap
+```
+
+Then, install npm packages required for the UI:
+
+```bash
+sudo npm install -g bower yarn webpack gulp
+```
+
+Start Memcached:
+
+```bash
+sudo systemctl enable memcached
+sudo systemctl start memcached
+```
+
+Then, configure, install and run PostgreSQL:
+
+```bash
+sudo su postgres -c "initdb --locale $LANG -E UTF8 -D '/var/lib/postgres/data'"
+sudo echo "local all all trust" >> /var/lib/postgres/pg_hba.conf
+sudo systemctl enable postgresql
+sudo systemctl start postgresql
+sudo su postgres -c "psql -c \"CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartvm'\""
+```
+
+More information on PostgreSQL in Arch Linux can be found on [ArchWiki](https://wiki.archlinux.org/index.php/PostgreSQL).
+
 ### OS X
 
 TODO.
@@ -204,8 +238,11 @@ echo "DONE."
 
 # Installing Docker
 
-Docker is necessary to run `hawkinit`, which we will cover later. To install
-it, first remove any currently installed version of Docker:
+Docker is necessary to run `hawkinit`, which we will cover later.
+
+### Fedora
+
+To install Docker, first remove any currently installed version of Docker:
 
 ```bash
 sudo dnf remove docker docker-common docker-selinux docker-engine-selinux docker-engine
@@ -244,6 +281,38 @@ And finally, to test if everything is working ok:
 ```bash
 docker run alpine /bin/echo OK
 ```
+
+### Arch Linux
+
+Arch Linux contains Docker CE in its repositories. Thanks to that, Docker installation
+is simple as:
+
+```bash
+sudo pacman -S docker docker-compose
+```
+
+If you want to run docker as a regular user you can add yourself to the `docker` group:
+
+```bash
+sudo usermod -a -G docker `whoami`
+newgrp docker # to update your groups on the session
+```
+Warning: Anyone added to the docker group is root equivalent. More information [here](https://github.com/docker/docker/issues/9976) and [here](http://docs.docker.com/engine/articles/security/).
+
+After that, enable and start Docker:
+
+```bash
+sudo systemctl enable docker
+sudo systemctl start docker
+```
+
+And finally, to test if everything is working ok:
+
+```bash
+docker run alpine /bin/echo OK
+```
+
+Advanced Docker configuration in Arch Linux is covered by [ArchWiki](https://wiki.archlinux.org/index.php/Docker).
 
 # Installing Hawkular
 
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -54,7 +54,7 @@ sudo su postgres -c "psql -c \"CREATE ROLE root SUPERUSER LOGIN PASSWORD 'smartv
 First, install all system dependencies:
 ```bash
 sudo pacman -S git memcached postgresql bzip2 libffi readline sqlite nodejs gcc
-# From AUR (i.e. via pacaur)
+# From AUR (e.g. via pacaur)
 pacaur -S openscap
 ```
 


tumido:
@abonas, would you suggest to leave the common parts for Fedora and Arch (like the Docker setup) at one place and link to it from the Arch section, or would you prefer to duplicate the paragraphs (as it is now in this PR) so it's easier to follow (no need to scroll back and forth) and also more robust to future changes?

abonas:
I think duplication is less preferable because then it is harder to maintain it and it's more work for same result.
but I am open to hear other opinions too. @cfcosta @karelhala @mtho11 @israel-hdez @xeviknal

abonas:
also, if we add more and more envs to the doc it becomes very long. if there's any way to make some sections collapse/expand, that would be great. I don't remember however if it's doable or not in this format.

xeviknal:
From the point of view of the user who is reading the manual, it is easier to have both Fedora/Arch sections completely split. Here an example of [ManageIQ guides](https://github.com/ManageIQ/guides/blob/master/developer_setup.md#install-system-packages)

In order to make it shorter and better readable, we could add an index bullet list with links to specific sections. Somehow like this: [Guides index](https://github.com/ManageiQ/guides).

tumido:
@xeviknal thanks for the examples! The first example was actually the reason why I've chosen to do it this way. On the other hand I have the some concern as @abonas about future updating and the need to cover and pay more attention to duplicates etc..

That being said, collapsible sections are [definitely possible](https://github.com/dear-github/dear-github/issues/166) with Github flavored Markdown. And it would look like this [Gist](https://gist.github.com/tumido/d9d63e4bd5574902133efad13d383d04).

israel-hdez:
I think that having sections for each distro is handy. However, IMO we only should include instructions for distros we can provide support directly and everything else should be on community forums or wikis. Our repo doesn't have a wiki, but we have ManageIQ Talk as the community forum.

cfcosta:
First of all, nicely done! More documentation is always good! That being said, I don't think we should add this to the guide unless we plan on supporting it for a long time. Since most of the people on the team work with Fedora, I think that means that we should support only Fedora and OS X.

If the other people disagree with me, then no problem, let's merge it!

abonas:
@israel-hdez @cfcosta I hear your concerns about supported/unsupported distros, it makes sense, so I agree we can certainly link to it and not make it part of the official guide.

cfcosta:
@abonas what do you mean by "link"? Put it in another markdown file on the repo or link to an external resource? That could be a good blog post or something like that.

Also, I think we should enable wiki on this repository and move all the documentation there.

mtho11:
We are not the first provider to document this kind of stuff, and being under the *ManageIQ org* we should strive to be like the other providers with look-n-feel of documentation. Perhaps under the [ManageIQ Guides](https://github.com/ManageIQ/guides) section is more appropriate and available to a wider audience of ManageIQ users.

IMO the current [Hawkular provider section](https://github.com/ManageIQ/guides/blob/master/providers/hawkular-hawkinit.md) in ManageIQ Guides is just focused on *HawkInit* which should be a sub-section of the more general *Hawkular Provider Setup*.

https://github.com/ManageIQ/guides/blob/master/providers/hawkular-hawkinit.md

To summarize, I think to be consistent with ManageIQ we should:

1. move our current [Hawkular setup Guide](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/GETTING_STARTED.md) over to [ManageIQ Guides](https://github.com/ManageIQ/guides). 

Or, 

2. link to our guide from the ManageIQ Guide page -- this might be best because we maintain control over the guide page content merging.

abonas:
> @abonas what do you mean by "link"? Put it in another markdown file on the repo or link to an external resource? That could be a good blog post or something like that.

does not really matter, I think it's easier inside the repo, as we don't have a relevant blog.
>
> Also, I think we should enable wiki on this repository and move all the documentation there.

no objections from my side

abonas:
> We are not the first provider to document this kind of stuff, and being under the ManageIQ org we should strive to be like the other providers with look-n-feel of documentation. Perhaps under the ManageIQ Guides section is more appropriate and available to a wider audience of ManageIQ users.

>
where is the similar documentation for developers for other providers/how does it look? 
 
> IMO the current Hawkular provider section in ManageIQ Guides is just focused on HawkInit which should be a sub-section of the more general Hawkular Provider Setup.
> 
> https://github.com/ManageIQ/guides/blob/master/providers/hawkular-hawkinit.md
> 
> To summarize, I think to be consistent with ManageIQ we should:
> 
> move our current Hawkular setup Guide over to ManageIQ Guides.

then we lose merge rights and the ability to quickly update what we need there, so I would not do that

> Or,
> 
> link to our guide from the ManageIQ Guide page -- this might be best because we maintain control over the guide page content merging.

I am not sure miq core page is the relevant place to link to our very specific provider, and there are many providers in miq, linking to all of them will expand the core dev guide and there's a relatively small team who eventually needs that hawkular guide. 
But you can certainly discuss it with core team first and see if they are interested in adding a link to our guide, if they are - that's great, you can send a PR with the link to our guide.

mtho11:
@abonas in answer to your question:

> where is the similar documentation for developers for other providers/how does it look?

From: https://github.com/ManageIQ/guides

- Provider setup instructions
    - [Amazon AWS](providers/amazon_aws_config.md)
    - [Hawkular](providers/hawkular-hawkinit.md)
    - [Openshift](providers/openshift.md)
    - [Openstack Infra](providers/openstack_infra_provider.md)

abonas:
> @abonas in answer to your question:
> 
> where is the similar documentation for developers for other providers/how does it look?
> From: https://github.com/ManageIQ/guides
> 
> Provider setup instructions
> Amazon AWS
> Hawkular
> Openshift
> Openstack Infra

@mtho11 I see a mismatch between that link that you mentioned and what is published on the manageiq website, it looks different:
I don't see those links in the website and I am unclear what is actually the correct one and mostly used in that case.
http://manageiq.org/docs/guides/developer_setup/

@agrare @blomquisg @jrafanie any idea?

jrafanie:
@abonas Good question.  I don't know.  I wonder if it's manually kept in sync and hasn't been done in a while.  You can ask in https://gitter.im/ManageIQ/manageiq.org.  

It looks like some of that was already brought up here:
https://github.com/ManageIQ/manageiq.org/issues/423

and the getting started guide has been updated, at least in github:
https://github.com/ManageIQ/manageiq.org/pull/588

tumido:
Do we have any resolution on this? Should I rather do a PR against [ManageIQ/guides](https://github.com/ManageIQ/guides) and modify the [developer_setup.md](https://github.com/ManageIQ/guides/blob/master/developer_setup.md)? I'm for unifying the guides or having them bundled somewhere. It seems to me, it's much better accessible and we're not duplicating our efforts. Also keeping all the docs in one repository would be more convenient for maintenance. Is there any possibility of gaining merge rights on [ManageIQ/guides](https://github.com/ManageIQ/guides)?

abonas:
@tumido I  prefer not to place our documentation in the main guide for several reasons:
1. our docs are not relevant for the majority of developers and can confuse them (not everyone should be forking providers-hawkular repo or api repo or starting hawkinit)
2. not having merge rights for our own documentation is very problematic. if we need something urgently updated (or even for onboarding more new developers), it is best to be able to do it quickly and not waiting for merging by another team, which from past experience can be extremely slow.

as noted above in previous comments: as some team members think - because archlinux is unsupported, it should be moved to a separated doc/wiki/whatever is convenient, and  linked from our guide that should remain here in this repo.

mtho11:
@abonas does this mean we should close this at this point?

tumido:
@mtho11 I would say we can close it, I'll write a blog post or a topic in the forum and that's about it. If there's no need to include this in the Hawkular guides, there's no need for having this PR left open/merged. I'll sync with core team on that matter since it's rather a ManageIQ specific guide than a Hawkular one and I'll see. But we can consider this PR closed. ;)

PR #59, opened 2017-09-26
Fixes event handling and generation of alternate machine in parser

israel-hdez:
* Event catcher was testing the wrong type and events weren't being passed to miq core for handling.
* alternate_machine_id method was assuming that the recevied machine id was 32 characters length and failing when this was not the case.
* Feed id was being escaped when adding members to Hawkular alert profiles and linking the alert to the wrong metrics in some cases. Escaping is removed because isn't needed as noted in https://github.com/ManageIQ/manageiq/issues/15756#issuecomment-324452842

Event handling issue: ManageIQ/manageiq#15756

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
@@ -105,7 +105,6 @@ module ManageIQ::Providers
       end
 
       def alternate_machine_id(machine_id)
-        return if machine_id.nil?
         # See the BZ #1294461 [1] for a more complete background.
         # Here, we'll try to adjust the machine ID to the format from that bug. We expect to get a string like
         # this: 2f68d133a4bc4c4bb19ecb47d344746c . For such string, we should return
@@ -113,6 +112,7 @@ module ManageIQ::Providers
         # providers store it in downcase, so, we let the upcase/downcase logic to other methods with more
         # business knowledge.
         # 1 - https://bugzilla.redhat.com/show_bug.cgi?id=1294461
+        return nil if machine_id.nil? || machine_id.length != 32 || machine_id[/\H/]
         alternate = []
         alternate << swap_part(machine_id[0, 8])
         alternate << swap_part(machine_id[8, 4])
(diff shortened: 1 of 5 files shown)


israel-hdez:
@miq-bot add_label bug, alerts, events, inventory

israel-hdez:
@miq-bot add_label wip

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -113,6 +113,7 @@ def alternate_machine_id(machine_id)
         # providers store it in downcase, so, we let the upcase/downcase logic to other methods with more
         # business knowledge.
         # 1 - https://bugzilla.redhat.com/show_bug.cgi?id=1294461
+        return nil if machine_id.nil? || machine_id.length != 32 || machine_id[/\H/]
```
abonas: it looks like there is already a check for machine_id.nil on line 108 ? and is the big explanation on lines 109 to 115 still stand?

israel-hdez: Oh! I didn't see that line. Thanks!
And I think that explanation still apply. We cannot be sure if the kernel of the machine is already patched. I think we should keep it just in case.

israel-hdez:
@miq-bot remove_label wip

abonas:
if this PR supposed to fix an issue, you can auto close the issue when it is merged by doing [this](https://help.github.com/articles/closing-issues-using-keywords/)

abonas:
@cfcosta please review

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
cfcosta: I think this would be better:

```ruby
return nil if machine_id.to_s.length != 32 || machine_id.to_s[/\H/]
```
Also, what does `\H` matches? That's a new one to me.

abonas:
Is this good to go? I saw it was approved but has a question at the same time

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
israel-hdez: From https://ruby-doc.org/core-2.2.0/Regexp.html:
> `/\H/` - A non-hexdigit character ([^0-9a-fA-F])

:)

cfcosta: Oh, nice!

israel-hdez:
I'd leave the commented line as is for now. Good to go for me.

cfcosta:
Yeah, it is.

PR #60, opened 2017-09-29
Migrate MiddlewareServer to MiddlewareServerWildfly and MiddlewareServerEap

tumido:
Sub-classing MiddlewareServer into more specific classes for Wildfly and EAP
 
Depends on: https://github.com/ManageIQ/manageiq-schema/pull/81
Depends on: https://github.com/ManageIQ/manageiq/pull/16126
Resolves: https://issues.jboss.org/browse/HAWKULAR-1248

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
@@ -332,12 +332,21 @@ module ManageIQ::Providers
           not_started && eap.properties[x].nil? ? _('not yet available') : eap.properties[x]
         end
 
-        inventory_object.assign_attributes(
+        attributes = {
           :name      => name || parse_standalone_server_name(eap.id),
           :type_path => eap.type_path,
           :hostname  => hostname,
           :product   => product
-        )
+        }
+
+        case product
+        when /wildfly/i
+          attributes[:type] = 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerWildfly'
+        when /eap/i
+          attributes[:type] = 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerEap'
+        end
+
+        inventory_object.assign_attributes(attributes)
       end
 
       def associate_with_vm(server, feed)
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server_eap.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server_eap.rb
@@ -1,0 +1,4 @@
+module ManageIQ::Providers
+  class Hawkular::MiddlewareManager::MiddlewareServerEap < Hawkular::MiddlewareManager::MiddlewareServerWildfly
+  end
+end
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server_wildfly.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server_wildfly.rb
@@ -1,21 +1,4 @@
 module ManageIQ::Providers
-  class Hawkular::MiddlewareManager::MiddlewareServerWildfly < MiddlewareServer
-    AVAIL_TYPE_ID = 'Server%20Availability~Server%20Availability'.freeze
-
-    has_many :middleware_diagnostic_reports, :dependent => :destroy
-
-    def feed
-      CGI.unescape(super)
-    end
-
-    def immutable?
-      properties['Immutable'] == 'true'
-    end
-
-    def enqueue_diagnostic_report(requesting_user:)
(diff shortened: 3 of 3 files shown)


tumido:
cc @cfcosta please review

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -332,12 +332,21 @@ def parse_middleware_server(eap, inventory_object, domain = false, name = nil)
...
           :product   => product
-        )
+        }
+
+        case product
+        when /wildfly/i
+          attributes[:type] = 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerWildfly'
+        when /eap/i
+          attributes[:type] = 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerEap'
+        end
```
Fryguy: - Minor, but you can extract the commonailty in the two branches like so:

```ruby
attributes[:type] =
  case product
  when /wildfly/i then 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerWildfly'
  when /eap/i    then 'ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServerEap'
  end

- What happens if the product is *neither* of those two cases for some reason.  Should you raise an exception?

**Comment by Fryguy** [comment]
@tumido I modified your OP to show that it "Depends on" the schema as opposed to just "Related to".  This ensure the merger knows the PR dependecny chain.

File: app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb (hunk 1 above)
Comment (tumido) [reply app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb]: I don't think that's needed, since we're 'just' selecting a more specific type. If neither case is matched, the `type` is left unchanged, which means the value is `ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServer`

**Comment by tumido** [comment]
@Fryguy thanks for the OP modification! :)

**Comment by tumido** [comment]
@abonas I know, yet my guide suggested to devide WildFly and EAP into two and having EAP to inherit from WildFly. @cfcosta can you please comment on that? I'm fine with both. There's been also some suggestions about the icon decorator to make it more robust in future and not to be fully dependent on product string matching each time the icon is displayed, etc. It would also justify to have WildFly and EAP as separate. But I'm not able to explain it properly, @cfcosta is the author of the idea....

**Comment by abonas** [comment]
@tumido @cfcosta please discuss this with @jdoyleoss, he might have some considerations regarding each solution. Whatever you agree on with him, is fine with me. If there are any changes in the requirement following your discussion, please document it in the JIRA properly.

**Comment by cfcosta** [comment]
@abonas @tumido @jdoyleoss The idea on separating the two classes is just an implementation detail, the idea is that if we do decide later on have something different for EAP, we would be able to do the changes without having to create a new migration.

**Comment by jdoyleoss** [comment]
@cfcosta If this an implementation, then It's in your domain to make that decision. My expectation is that the features available to users of WildFly and EAP are the same. We need different icons and possibly different descriptive text that might not come from the server itself. As long as we can support non-functional differences like that, we're likely OK.

**Comment by abonas** [comment]
adding to @jdoyleoss comment, on implementation side, lets also avoid copy pasting code in 2 different classes. (that is - if anything specific is needed in a similar way in both, for example, for power operations, calculating status, or alerts/metrics, etc.)

**Comment by tumido** [comment]
Added a PR against core https://github.com/ManageIQ/manageiq/pull/16126 so the type can be propagated.

@abonas, yes that's also possible with this implementation. WildFly is a parent class for EAP, so any downstream specific code can be kept separate, yet all common parts for WildFly and EAP can be placed into WildFly class, and EAP will inherit it.

**Comment by tumido** [comment]
@cfcosta can you please approve this PR, if there are no concerns left to discuss?

**Comment by cfcosta** [comment]
I think this is ok to me.

File: app/models/manageiq/providers/hawkular/middleware_manager/middleware_server_eap.rb [hunk 2]
```diff
@@ -0,0 +1,4 @@
+module ManageIQ::Providers
+  class Hawkular::MiddlewareManager::MiddlewareServerEap < Hawkular::MiddlewareManager::MiddlewareServerWildfly
```
abonas: is based on this hierarchy eap a subclass of wildfly? if yes - why?

abonas: the requirement is that there is a general middleware_server, and it has subclasses such as eap, wildfly and in the future Fuse and others

tumido:
@abonas, can you please approve the PR now?

Fryguy:
Merged the migration, so merging this in tandem.

PR #61, opened 2017-10-02
Support for timeline events for MW Server and MW domain power operations

israel-hdez:
When server or domain operation is done, an event in the timeline
is logged. Either if the operation succeds of fails.

Jira: https://issues.jboss.org/browse/HAWKULAR-1249

Must be merged first: ManageIQ/manageiq-schema#86

screenshot from 2017-10-04 12-42-09

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -139,42 +139,42 @@ module ManageIQ::Providers
       run_generic_operation(:Shutdown, ems_ref, :restart => false, :timeout => timeout)
     end
 
-    def suspend_middleware_server(ems_ref, params = {})
+    def suspend_middleware_server(ems_ref, params = {}, extra_data = {})
       timeout = params[:timeout] || 0
-      run_generic_operation(:Suspend, ems_ref, :timeout => timeout)
+      run_generic_operation(:Suspend, ems_ref, {:timeout => timeout}, extra_data)
     end
 
-    def resume_middleware_server(ems_ref)
-      run_generic_operation(:Resume, ems_ref)
+    def resume_middleware_server(ems_ref, extra_data = {})
+      run_generic_operation(:Resume, ems_ref, {}, extra_data)
     end
 
-    def reload_middleware_server(ems_ref)
-      run_generic_operation(:Reload, ems_ref)
+    def reload_middleware_server(ems_ref, extra_data = {})
+      run_generic_operation(:Reload, ems_ref, {}, extra_data)
     end
 
     def stop_middleware_server(ems_ref)
-      run_generic_operation(:Shutdown, ems_ref)
+      run_generic_operation(:Shutdown, ems_ref, {}, {:original_operation => :Stop})
     end
 
-    def start_middleware_domain_server(ems_ref)
-      run_generic_operation(:Start, ems_ref)
+    def start_middleware_domain_server(ems_ref, extra_data = {})
+      run_generic_operation(:Start, ems_ref, {}, extra_data)
     end
 
-    def stop_middleware_domain_server(ems_ref)
-      run_generic_operation(:Stop, ems_ref)
+    def stop_middleware_domain_server(ems_ref, extra_data = {})
+      run_generic_operation(:Stop, ems_ref, {}, extra_data)
     end
 
     def restart_middleware_server(ems_ref)
-      run_generic_operation(:Shutdown, ems_ref, :restart => true)
+      run_generic_operation(:Shutdown, ems_ref, {:restart => true}, {:original_operation => :Restart})
     end
 
     # domain server ops
-    def restart_middleware_domain_server(ems_ref)
-      run_generic_operation(:Restart, ems_ref)
+    def restart_middleware_domain_server(ems_ref, extra_data = {})
+      run_generic_operation(:Restart, ems_ref, {}, extra_data)
     end
 
-    def kill_middleware_domain_server(ems_ref)
-      run_generic_operation(:Kill, ems_ref)
+    def kill_middleware_domain_server(ems_ref, extra_data = {})
+      run_generic_operation(:Kill, ems_ref, {}, extra_data)
     end
 
     # server group ops
@@ -234,7 +234,7 @@ module ManageIQ::Providers
           on.failure do |error|
             _log.error 'error callback was called, reason: ' + error.to_s
             emit_middleware_notification(:mw_op_failure, 'Add Datasource', datasource_data[:datasourceName], ems_ref,
-                                         MiddlewareServer)
+                                         MiddlewareServer, error.to_s)
           end
         end
       end
@@ -267,7 +267,7 @@ module ManageIQ::Providers
           on.failure do |error|
             _log.error 'error callback was called, reason: ' + error.to_s
             emit_middleware_notification(:mw_op_failure, 'Deploy', deployment_data[:destination_file_name], ems_ref,
-                                         MiddlewareServer)
+                                         MiddlewareServer, error.to_s)
           end
         end
       end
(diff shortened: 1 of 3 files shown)


israel-hdez:

israel-hdez:
@miq-bot add_label wip

israel-hdez:
@miq-bot remove_label wip

israel-hdez:
@cfcosta I hope you can give this a look.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -498,18 +525,27 @@ def run_specific_operation(operation_name, ems_ref, parameters = {})
...
 
-    def run_operation(parameters, operation_name = nil)
+    def run_operation(parameters, operation_name = nil, extra_data = {})
       with_provider_connection do |connection|
         callback = proc do |on|
           on.success do |data|
-            _log.debug "Success on websocket-operation #{data}"
-            emit_middleware_notification(:mw_op_success, parameters[:operationName], nil, parameters[:resourcePath],
-                                         MiddlewareServer)
+            emit_middleware_notification(
```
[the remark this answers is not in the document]
israel-hdez: I was thinking in changing `emit_middleware_notification` signature to named parameters. wdyt? may that be enough?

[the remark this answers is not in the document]
cfcosta: That would be good, but even better would be something like this, imho: https://refactoring.guru/introduce-parameter-object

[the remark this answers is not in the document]
israel-hdez: Hmm... I haven't seen the parameter object pattern widely used in ruby world. But I can see the benefits for dry-ing up a little. So, I applied it. Please check again. I hope this is what you expected.

israel-hdez:
CI is now passing.
@cfcosta @mtho11 , are you OK with the warnings? If so, I think this could be merged once @cfcosta approves it.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -498,18 +533,24 @@ def run_specific_operation(operation_name, ems_ref, parameters = {})
...
 
-    def run_operation(parameters, operation_name = nil)
+    def run_operation(parameters, operation_name = nil, extra_data = {})
       with_provider_connection do |connection|
         callback = proc do |on|
-          on.success do |data|
-            _log.debug "Success on websocket-operation #{data}"
-            emit_middleware_notification(:mw_op_success, parameters[:operationName], nil, parameters[:resourcePath],
-                                         MiddlewareServer)
+          notification_args = EmitNotificationArgs.new(
```
cfcosta: This is good, but all the responsibilities on mapping the data correctly are still on the caller. `extra_data` and `parameters` could be sent to the object directly, and it would get the correct key from there. I've been thinking something like this would be a good interface, which would move the two conditional and data introspection to the data object:

```ruby
EmitNotificationArgs.new(:mw_op_success, MiddlewareServer, parameters, extra_data)
```

Also, for the name, what do you think of just `NotificationArgs`?

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -44,4 +44,141 @@
...
+    def timeline_domain_expectations(status)
+      event = EmsEvent.last
+      expect(event.ems_id).to eq(ems.id)
+      expect(event.event_type).to eq("MwDomain.Stop.#{status}")
+      expect(event.middleware_domain_id).to eq(mw_domain.id)
+      expect(event.middleware_domain_name).to eq(mw_domain.name)
+    end
+
+    %w(shutdown suspend resume reload restart stop).each do |operation|
+      it "#{operation} should create a user notification and timeline event on success" do
```
cfcosta: This can be collapsed into a `describe` block, for something like this:
```ruby
describe operation do
  it 'should .*"
end
```

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -44,4 +44,141 @@
...
+        MiqQueue.last.deliver
+
+        op_name = operation.capitalize
+        notification_expectations(mw_server, op_name.to_sym, 'mw_op_failure')
+        timeline_server_expectations(mw_server, op_name, "Failed")
+      end
+    end
+
+    %w(start restart stop kill).each do |operation|
+      it "domain server specific '#{operation}' operation should create a user notification and timeline event on success" do
```
cfcosta: same here:
```ruby
describe "domain server specific: '#{operation}'"
```

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -521,6 +562,54 @@ def run_operation(parameters, operation_name = nil)
...
+        EmsEvent.add_queue(
+          'add', id,
+          :ems_id          => id,
+          :source          => 'HAWKULAR',
+          :timestamp       => Time.zone.now,
+          :event_type      => '%{entity_type}.%{operation}.%{status}' %
+            { :entity_type => mw_server.kind_of?(MiddlewareServer) ? 'MwServer' : 'MwDomain',
+              :operation   => notification_args.operation_name,
+              :status      => notification_args.type == :mw_op_success ? 'Success' : 'Failed' },
+          :message         => message,
```
cfcosta: This could also be a method in `EmitNotificationArgs` itself.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -521,6 +562,54 @@ def run_operation(parameters, operation_name = nil)
...
+                    :status    => notification_args.type == :mw_op_success ? _('succeeded') : _('failed')
+                  }
+        message += ": #{notification_args.detailed_message}" if notification_args.detailed_message
+
+        EmsEvent.add_queue(
+          'add', id,
+          :ems_id          => id,
+          :source          => 'HAWKULAR',
+          :timestamp       => Time.zone.now,
+          :event_type      => '%{entity_type}.%{operation}.%{status}' %
```
cfcosta: This could be a method in EmitNotificationArgs itself. The idea behind the args object is to not query the data directly and then decide how to act on it, but to get the final data that we do need, and not introspect on the object itself. So something like that:

```ruby
notification_args.event_type
```

app/models/manageiq/providers/hawkular/middleware_manager.rb
israel-hdez: Ok, I'll use `NotificationArgs`.
I'll skip the suggestion of moving the mapping responsibility, because this is the only usage where it makes sense and breaks all other usages.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -521,6 +564,60 @@ def run_operation(parameters, operation_name = nil)
       end
     end
 
+    NotificationArgs = Struct.new(:type, :operation_name, :operation_args,
+                                  :target_resource, :entity_klass, :detailed_message) do
+      def event_type(mw_entity)
+        '%{entity_type}.%{operation}.%{status}' %
+          { :entity_type => mw_entity.kind_of?(MiddlewareServer) ? 'MwServer' : 'MwDomain',
```
cfcosta: I think this is `entity_klass`, no?

israel-hdez: You made me doubt for a moment. But it's ok to check `mw_entity`. `entity_klass` can be, for example, a `MiddlewareDatasource`, and `mw_entity` will be the owning `MiddlewareServer` server.

israel-hdez:
@mtho11 For my side I think it's ready. Any other thoughts, @cfcosta ?
Looks like there is a recently injected bug in `EmsEvent` that is preventing lots of our test to pass. What do we do? Do you wan't to merge regardless that?

israel-hdez:
CI is now green @mtho11

PR #71, opened 2017-10-06
Create EAP operation Events

tumido:
Add events for:

- [x] EAP Application Deployment
- [x] EAP Application Undeplyment
- [x] EAP JDBC Driver Deployment
- [x] EAP JDBC Driver Undeployment
- [x] EAP Datasource Creation
- [x] EAP Datasource Deletion

Related to:
- https://github.com/ManageIQ/manageiq/pull/16134
- https://github.com/ManageIQ/manageiq-ui-classic/pull/2317
- https://github.com/ManageIQ/manageiq-ui-classic/pull/2331
- https://github.com/hawkular/hawkular-services/pull/191

Resolves:
- https://issues.jboss.org/browse/JMAN4-200

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -10,12 +10,16 @@
           - hawkular_datasource_remove.error
           - hawkular_deployment.error
           - hawkular_deployment_remove.error
+          - hawkular_jdbc.error
+          - hawkular_jdbc_remove.error
           - hawkular_event.critical
           :detail:
           - hawkular_datasource.ok
           - hawkular_datasource_remove.ok
           - hawkular_deployment.ok
           - hawkular_deployment_remove.ok
+          - hawkular_jdbc.ok
+          - hawkular_jdbc_remove.ok
           - hawkular_event
     :jdr:
       :generation_timeout: 3.minutes


tumido:
@israel-hdez, can you help me, please?

Is this the proper way, how to list such events? Also I'm not sure what to do about the mapping in [manageiq-content/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class](https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class) dir... (I don't want to duplicate the events, but I'm not sure if the provider or ManageIQ should be firing them).

Also, how should I model the events for Application (un)deployment and JDBC (un)deployment? Are these the right places?
- [add_middleware_deployment](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/0b29da92a469077e9968ff88ff42340ce18eb2c5/app/models/manageiq/providers/hawkular/middleware_manager.rb#L243)
- [undeploy_middleware_deployment](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/0b29da92a469077e9968ff88ff42340ce18eb2c5/app/models/manageiq/providers/hawkular/middleware_manager.rb#L276)
 - [add_middleware_jdbc_driver](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/0b29da92a469077e9968ff88ff42340ce18eb2c5/app/models/manageiq/providers/hawkular/middleware_manager.rb#L363)

Or should I be adding them on `Hawkular::MiddlewareManager::MiddlewareDeployment` creation and deletion or in [manageiq-ui-classic/app/controllers/middleware_server_controller.rb](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/middleware_server_controller.rb)? And do we have any hook for JDBC undeployment? I struggle to find any... :/

Too many questions I know.. :P

tumido:
@miq-bot add_label events, enhancement

israel-hdez:
@tumido The repo `manageiq-content` is for the _Automate_ feature. I think we don't need any changes there for now.

I've checked your ManageIQ/manageiq#16134 and seems fine. Probably, the only thing that you should change is that the event identifiers (the names) should match the ones we have [here](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/0b29da92a469077e9968ff88ff42340ce18eb2c5/config/settings.yml#L9-L19). I'm almost sure that if they don't match, any configurations done with the events won't fire.

I've read multiple times the Jira, and I don't fully understand if it's only required to list the events in _Control Explorer_, or if also the timeline events are required. I'm confused because in JIRA, the _user story_ field mentions the timeline, but in the _description_ it's only mentioned the _control explorer_. So, before going forward, I'd suggest you to ask in JIRA for clarification (tagging John).

tumido:
@israel-hdez Hm, ok, I'll change the names then. But I'm not much convinced about the JDBC events. From the list you've showed me, it seems there are no events for JDBC drivers yet. Is that correct? Should I add them in the settings as well?

It also seems like the events `hawkular_datasource...` are never fired in ManageIQ and `hawkular_deployment...` are fired only in the `manageiq-content`. Which leads me to thinking the firing should be added by me, yet I'm not sure when/where.

israel-hdez:
@tumido The datasource and deployment events happen in WildFly Agent and are made available to MiQ through Hawkular services. If you query your Hawkular instance at http://hawkular:8080/hawkular/alerts/events (or whichever host/port you have it) you should be able to see them (if there are any). These events are collected in the provider's [event catcher](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/2720b9e636fa754f5a5ebb4d529305afc6fda2df/app/models/manageiq/providers/hawkular/middleware_manager/event_catcher/stream.rb#L40). Once collected, the events are passed to MiQ core for processing to finally end stored in `event_streams` which is the table the timeline fetches its data from.

Hmm... I can see that there are no JDBC events in `config.yml`. Perhaps you can try adding a JDBC driver to one of your servers to see if an event appears in Hawkular? If it does, then only adding the required lines to the config file will make the events show in the timeline. If not, we will need help from somebody else to make WildFly Agent push to Hawkular the required events.

tumido:
I've changed the names on events so it's matching the `settings.yml` and the names used by Hawkular. I've also discovered that there are **no JDBC related events** present in Hawkular by now.

tumido:
@miq-bot remove_label wip

tumido:
@israel-hdez, can you please review this? Is there any testsuite testing events which I should adjust?

israel-hdez:
Looks fine :+1:

abonas:
@tumido as there are many PRs for this feature, what is the correct order/dependency between them? especially since they are in different repos and don't "build" on top of each other in a single repo

tumido:
@abonas, chained from the first to last:

1. We need the https://github.com/hawkular/hawkular-services/pull/191 to expose the `hawkular_jdbc` events
2. The https://github.com/ManageIQ/manageiq-providers-hawkular/pull/71 catches said events in ManageIQ
3. The https://github.com/ManageIQ/manageiq/pull/16134 shows all 6 event types in the `Control -> Explorer -> Events`
4. https://github.com/ManageIQ/manageiq-ui-classic/pull/2317 and https://github.com/ManageIQ/manageiq-ui-classic/pull/2331 assign Middleware icons in the `Control -> Explorer -> Events` (without these PRs there's the default star icon shown)

Without any of the PRs with lower number the later has no sense and the code is redundant but nothing will fail.
Example:
- Without no. 2 the 3rd will display all the events but the JDBC ones are never propagated into ManageIQ (never catched).
- Without no 1 - ManageIQ is prepared to display any events but JDBC events are not propagated. etc.
- Without 3rd the PRs in 4 has no sense but nothing is broken.

It's not dependent in implementation, they are complementary.

israel-hdez:
@tumido This PR can be merged, right?

tumido:
@israel-hdez yes! 🚀

israel-hdez:
@miq-bot assign @mtho11 
Looks good to go.

PR #72, opened 2017-10-06
Use base class MiddlewareDiagnosticReport so we can fully use report to display data

karelhala:
### Merge after https://github.com/ManageIQ/manageiq/pull/16135
As stated in PR for core repo, we have to use different base class otherwise we are unable to fully use report data and display data in table with pagination, sort and such.

While changing the base class I also altered the class MiddlewareDiagnosticReport not to include table name and belongs to, which should be part of core functionality.


abonas:
@karelhala please rekick travis, I don't have permissions, and for some reason reopening PR did not do that automatically (it does so in other repos I worked with)

abonas:
LGTM but @israel-hdez is the main reviewer here, please review

PR #74, opened 2017-10-12
[WIP] Implement generic view mapping

cfcosta:
This is a first step into actually implementing the proof of concept into proper ManageIQ. Currently it enables rendering as json of data imported from hawkular. It is not supposed to be merged as-is, but used as a base branch to generate UI code upon (@mtho11).

To test:

```ruby
resources = ManageIQ::Providers::Hawkular::Connection.new.fetch_all_resources
```

All generated view objects respond to #as_json, hence something like
this on a controller would render all captured objects:

```ruby
def index
  render json: ManageIQ::Providers::Hawkular::Connection.new.fetch_all_resources
```

### DIFF
--- a/lib/manageiq-providers-hawkular.rb
+++ b/lib/manageiq-providers-hawkular.rb
@@ -1,2 +1,12 @@
-require "manageiq/providers/hawkular/engine"
-require "manageiq/providers/hawkular/version"
+require 'manageiq/providers/hawkular/engine'
+require 'manageiq/providers/hawkular/version'
+
+require 'generic_view_mapper'
+
+require 'manageiq/providers/hawkular/connection'
+require 'manageiq/providers/hawkular/resource_collection'
+
+root = File.expand_path('./manageiq/providers/hawkular', __dir__)
+Dir.glob(File.join(root, '{entities,views}', '*.rb')) do |file|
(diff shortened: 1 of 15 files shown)


xeviknal:
LGTM! :+1: 

I understand from this PR that all kinds of resources of the inventory are going to be rendered following `ManageIQ::Providers::Hawkular::Views::MiddlewareResourceView` except all those which are identified (as an entity) as WildflyServer. Those ones are going to follow the rules of `ManageIQ::Providers::Hawkular::Views::WildflyServerView`. Am I wrong?

In addition, I have a question here, is it useful to have a `Entity` which doesn't have a View *bound to* it? In this WIP PR, you have, for example, `ManageIQ::Providers::Hawkular::Entities::OperatingSystem` but is not bound to any specific view (except the `MiddlewareResourceView` from inheritance). Which responsability would have in the system this OperationSystem Entity?

cfcosta:
@xeviknal none, it would render directly to `MiddlewareResourceView`. The entities are created, in this case, only for mental mapping purposes, and they are not exactly required.

abonas:
@ManageIQ/committers-hawkular I'm testing the maintainers team notification, please ignore :)

PR #76, opened 2017-10-13
Explicitly pick from the connection pool for notifications

cfcosta:
Every time hawkular does some action into the server, it is run through
a different thread that connects to the server, and runs a callback when
the result is ready, on the same thread as the connection. After that,
it keeps the thread running for new requests.

As such, we now force the connection to be picked explicitly when
creating the notification, forcing ActiveRecord to release the
connection after the callback is run.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1420431

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -586,35 +586,37 @@ module ManageIQ::Providers
     private_constant :NotificationArgs
 
     def emit_middleware_notification(notification_args)
-      mw_entity = notification_args.entity_klass.find_by(:ems_ref => ems_ref) unless notification_args.entity_klass == MiddlewareServer
-      mw_server = if mw_entity.nil?
-                    MiddlewareServer.find_by(:ems_ref => notification_args.target_resource) ||
-                      MiddlewareDomain.find_by(:ems_ref => notification_args.target_resource)
-                  else
-                    MiddlewareServer.find_by(:id => mw_entity.server_id)
-                  end
+      ActiveRecord::Base.connection_pool.with_connection do
+        mw_entity = notification_args.entity_klass.find_by(:ems_ref => ems_ref) unless notification_args.entity_klass == MiddlewareServer
+        mw_server = if mw_entity.nil?
+                      MiddlewareServer.find_by(:ems_ref => notification_args.target_resource) ||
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -586,35 +586,37 @@ def event_message(mw_entity)
...
-        EmsEvent.add_queue(
-          'add', id,
-          :ems_id          => id,
-          :source          => 'HAWKULAR',
-          :timestamp       => Time.zone.now,
-          :event_type      => notification_args.event_type(mw_server),
-          :message         => notification_args.event_message(mw_server),
-          :middleware_ref  => mw_server.ems_ref,
-          :middleware_type => mw_server.class.name.demodulize
+      ActiveRecord::Base.connection_pool.with_connection do
```
Fryguy: @jrafanie Please review.

jrafanie: Make sense to me.  Previously, if the thread is used to checkout out a connection to do the notification but then persists to do other things, activerecord won't steal the connection back from that thread.  Any other similar requests that come into puma will consume the database connection pool.   This is exactly what I was seeing when analyzing the logs.

This looks like it should fix the problem, provided this long running thread does no other activerecord database "work" before or after this block.

jrafanie: @cfcosta looks great.  I'm good with merging this and releasing it if that's the only way to reasonably test it.

cfcosta: @jrafanie great! There's no other ActiveRecord work being done besides generating notifications, so this will fix the problem.

mtho11:
Gotta love @skateman **Approve** logo

mtho11:
Given the comments today, I will merge.

mtho11:
@abonas do you agree at this point in the sprint/lifecycle to merge?

abonas:
@mtho11 no objections. 
But as this was not merged yesterday, that means it is already entering next sprint afaik anyway, not 71.
please add milestones only when merging, otherwise it enters the milestone stats without actually being there :)

jrafanie:
Thanks @cfcosta 👍

cfcosta:
@jrafanie :3

PR #80, opened 2017-10-17
Fix min Ruby to be 2.3.1 per the recent core change, otherwise it fails

abonas:
Running with Ruby 2.3.0 gives this error:
There was an error parsing `Gemfile`: Ruby versions less than 2.3.1 are unsupported!. Bundler cannot continue.


abonas:
@mtho11 merge?

PR #92, opened 2017-10-31
Typo fixed on Live metrics name: mw_aggregated_servlet_time

xeviknal:
It is fixed a typo on a live metric name: mw_aggregated_servlet_time
Here is the fix on the core repo: https://github.com/ManageIQ/manageiq/pull/16363

@abonas who should I ping to review this? 
And how this dependency with core repo is handled? It needs some sync with core.

### DIFF
--- a/spec/models/manageiq/providers/hawkular/middleware_manager/middleware_server_spec.rb
+++ b/spec/models/manageiq/providers/hawkular/middleware_manager/middleware_server_spec.rb
@@ -37,7 +37,7 @@ describe ManageIQ::Providers::Hawkular::MiddlewareManager::MiddlewareServer do
       "WildFly Memory Metrics~NonHeap Used"                               => "mw_non_heap_used",
       "WildFly Memory Metrics~NonHeap Committed"                          => "mw_non_heap_committed",
       "WildFly Memory Metrics~Accumulated GC Duration"                    => "mw_accumulated_gc_duration",
-      "WildFly Aggregated Web Metrics~Aggregated Servlet Request Time"    => "mw_agregated_servlet_time",
+      "WildFly Aggregated Web Metrics~Aggregated Servlet Request Time"    => "mw_aggregated_servlet_time",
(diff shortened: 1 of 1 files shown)


abonas:
is there an open bugzilla/issue associated with this?

xeviknal:
No, there isn't. Actually it is not a bug because is not causing any problem or misbehaving. But it can lead some misleadings.

abonas:
> No, there isn't. Actually it is not a bug because is not causing any problem or misbehaving. But it can lead some misleadings.

so what is the implication if there is no bug/no misbehaving?

xeviknal:
It was a misleading on previous PR because the typo made people get confused. Also because I thought it could be prettier to have the word well-spelled. As well, it would follow the other `aggregated` metrics which are well-spelled. 

If you think it is no proper for now, let's close it.

abonas:
I am +1 for fixing this. 
isn't this influencing collecting/displaying the wrong metric though?

xeviknal:
The metric is not incorrect. What is incorrect is the mapping name (used for MIQ internal purposes). Then all the usages of this mapping have to be replaced in both of the apps: core and provider. 
In core repo, there is no usage of that metric. Which it makes sense since core should be provider agnostic.
In provider repo (here), I just found an occurrence. It is test making sure this metric is present and properly mapped. But I don't know if there is missing usages of this mapping somewhere.

xeviknal:
On the other side, tests are going to fail until the core PR will be merged.

simaishi:

```
$ git log -1
commit d6737a38af411d8029191cdb7bc92869fd90a643
Author: abonas <mikeyteva@gmail.com>
Date:   Wed Nov 1 11:11:53 2017 +0200

    Merge pull request #92 from xeviknal/typo-hawkular-metrics
    
    Typo fixed on Live metrics name: mw_aggregated_servlet_time
    (cherry picked from commit e5323c295753739dd4d110a6cb76fa985d70fcf3)
```

PR #100, opened 2017-11-06
Add optional validate param to raw_connect

tumido:
When provider is added the credentials validation is newly done via `raw_connect` (class method) instead of `verify_credentials` (needed EMS object). Further reading on this change can be found here: https://github.com/ManageIQ/manageiq-ui-classic/pull/1580

Related to: https://github.com/ManageIQ/manageiq-ui-classic/pull/2577

Similar PRs for other providers:
- https://github.com/ManageIQ/manageiq-providers-azure/pull/161
- https://github.com/ManageIQ/manageiq-providers-google/pull/28
- https://github.com/ManageIQ/manageiq-providers-scvmm/pull/41
- https://github.com/ManageIQ/manageiq-providers-lenovo/pull/108
- https://github.com/ManageIQ/manageiq-providers-nuage/pull/40

**This PR enables the new workflow on provider. With this PR both workflows are available.**

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -78,6 +78,7 @@ module ManageIQ::Providers
       $log.error(err)
       raise MiqException::Error, 'Unable to verify credentials'
     end
+    private_class_method :connection_rescue_block
 
     def validate_authentication_status
       {:available => true, :message => nil}
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -51,24 +51,32 @@ module ManageIQ::Providers
     attr_accessor :client
 
     def verify_credentials(_auth_type = nil, options = {})
-      begin
-        # As the connect will only give a handle
-        # we verify the credentials via an actual operation
-        connect(options).inventory.list_feeds
-      rescue URI::InvalidComponentError
-        raise MiqException::MiqHostError, "Host '#{hostname}' is invalid"
-      rescue ::Hawkular::ConnectionException
-        raise MiqException::MiqUnreachableError, "Unable to connect to #{hostname}:#{port}"
-      rescue ::Hawkular::Exception => he
-        raise MiqException::MiqInvalidCredentialsError, 'Invalid credentials' if he.status_code == 401
-        raise MiqException::MiqHostError, 'Hawkular not found on host' if he.status_code == 404
-        raise MiqException::MiqCommunicationsError, he.message
-      rescue => err
-        $log.error(err)
-        raise MiqException::Error, 'Unable to verify credentials'
-      end
+      self.class.validate_connection(connect(options), options[:hostname], options[:port])
 
       true
+    end
+
+    def self.validate_connection(connection, hostname = nil, port = nil)
+      # As the connect will only give a handle
+      # we verify the credentials via an actual operation
+      connection_rescue_block(hostname, port) do
+        connection.inventory.list_feeds
+      end
+    end
+
+    def self.connection_rescue_block(hostname = nil, port = nil)
+      yield
+    rescue URI::InvalidComponentError
+      raise MiqException::MiqHostError, "Host '#{hostname}' is invalid"
+    rescue ::Hawkular::ConnectionException
+      raise MiqException::MiqUnreachableError, "Unable to connect to #{hostname}:#{port}"
+    rescue ::Hawkular::Exception => he
+      raise MiqException::MiqInvalidCredentialsError, 'Invalid credentials' if he.status_code == 401
+      raise MiqException::MiqHostError, 'Hawkular not found on host' if he.status_code == 404
+      raise MiqException::MiqCommunicationsError, he.message
+    rescue => err
+      $log.error(err)
+      raise MiqException::Error, 'Unable to verify credentials'
     end
 
     def validate_authentication_status
@@ -94,7 +102,7 @@ module ManageIQ::Providers
     end
 
     # Hawkular Client
-    def self.raw_connect(host, port, username, password, security_protocol, cert_store)
+    def self.raw_connect(host, port, username, password, security_protocol, cert_store, validate = false)
       credentials = {
         :username => username,
         :password => password
@@ -104,8 +112,12 @@ module ManageIQ::Providers
         :verify_ssl     => verify_ssl_mode(security_protocol),
         :ssl_cert_store => cert_store
       }
-      ::Hawkular::Client.new(:entrypoint => entrypoint(host, port, security_protocol),
-                             :credentials => credentials, :options => options)
+      connection = ::Hawkular::Client.new(:entrypoint => entrypoint(host, port, security_protocol),
+                                          :credentials => credentials, :options => options)
+
+      validate_connection(connection, host, port) if validate
+
+      connection
     end
 
     def connect(_options = {})


tumido:
cc @aljesusg, @abonas, @jntullo, please review

tumido:
@miq-bot add_label gaprindashvili/yes bug

abonas:
@agrare @cben  I can't add for some reason @jntullo as a reviewer. any idea why?

aljesusg:
With Correct credentials should be correct **I am getting Invalid Credentials**
should_be_correct

With incorrect credentials go OK
ok_other_credential

With a Incorrect Host go OK
ok_host_random

tumido:
@aljesusg it's not working because the mapping in your https://github.com/ManageIQ/manageiq-ui-classic/pull/2577 is sending `password` instead of `params[:default_password]`, when you fix it, it's gonna work.

aljesusg:
@tumido in get_task_args the first line is
**user, password = params[:default_userid], MiqPassword.encrypt(params[:default_password])**
So are we working without encryption password?

tumido:
@aljesusg I know.. It seems so. Without encryption it works for me. But I'm not sure how it behaves when the ssl is enabled, I haven't checked that.

aljesusg:
Ok I tested and work without encryption I am gonna test this with SSL and CA @tumido

aljesusg:
This fail with SSL 
**Credential validation was not successful: wrong argument (String)! (Expected kind of OpenSSL::X509::Store)**  thisis running in local with curl maybe we are missing something, I tested wwith encrypted password and is failing too, after the call I'll debug this.

credentials object
```ruby
{:username=>"jdoe", :password=>"password"}
```
options object
```ruby
{:tenant=>"hawkular", :verify_ssl=>1, :ssl_cert_store=>"-----BEGIN CERTIFICATE-----\r\nMIIC/TCCAeWgAwIBAgIEZi3DSDANBgkqhkiG9w0BAQsFADAcMRowGAYDVQQDExFo\r\nYXdrdWxhci1zZXJ2aWNlczAgFw0xNzExMDIwOTQxMzVaGA8yMTE3MTAwOTA5NDEz\r\nNVowHDEaMBgGA1UEAxMRaGF3a3VsYXItc2VydmljZXMwggEiMA0GCSqGSIb3DQEB\r\nAQUAA4IBDwAwggEKAoIBAQCYWbkxaDx6e9mMK/T87QSJ6NS5KtlPzx3RItmegeOs\r\n8Z9vxHMC6rDKr+qlX3Ed/OhIk36az8rSrOYHT3ONdVVOK/em0JnOq8+lFyL1sfDT\r\nnN63s3ON4p8JS1jUxVcIafTJu12h//xf/nVbW5q3pcE242AXL4q0LIDIGczm2cV5\r\nYjeofTLJl/ra+8+yAygUm6+kT1MTkzb8TtNiJJDxAUTDDBF+iuISV4R5adLGmCmz\r\njiDZPxiwHI4TLO0I02jIzpQ9Rql+7Sd7jnEzea9oFJT7dr3E+CbknvUO1bso+Jwt\r\ns7T9CeYzKxBswW83pPTI44iYmydtWwenzPqUivMIwio3AgMBAAGjRTBDMCIGA1Ud\r\nEQQbMBmHBH8AAAGCEWhhd2t1bGFyLXNlcnZpY2VzMB0GA1UdDgQWBBTfPluQkZHh\r\nCsKExlfsrrzpNhJpNzANBgkqhkiG9w0BAQsFAAOCAQEAM5MDnDcgz5XKc6FL/fBr\r\nQ4TIgNG9T7qFn2lLEkOkqh4kDAKfyFa5bBRTrDRglQoGQBfdAJD4nnS/mxBsLrBD\r\neW8bS9+1aGIyUnZK8fYIVR7wXSvb3U/+D0foqJJDEtyBftSsscrL1dYNEaOMrhAO\r\nECj7xl4638zdjp3klxlcIvJ9NINdPON4ltr796nhZIoHs8OTDW04M3sztzdpMkiH\r\ncBz1seRQlPAmWSK29y1fpXNrr0PmlYijgMYVv59o7zm4/Gvv2PcTzWh9ykFK/WnZ\r\nY1MooB0peBI4X3vWx/VAuX1ybUyPthx/gzbtZks6ztVjJOgs50eawUz3Q7S3BtWv\r\n5g==\r\n-----END CERTIFICATE-----"}

```

tumido:
@aljesusg ok, do you think this should be covered in this PR? Is it caused by me or just exposed?

josejulio:
I think you should take at the PR[1] that did that work.
The certificate is parsed to a `OpenSSL::X509::Store` by the method: `ssl_cert_store` 

[1] https://github.com/ManageIQ/manageiq/pull/13721/files#diff-5d3f1a5ab701ca4a756d3c921e36e9f6R22

aljesusg:
So why is getting this error @josejulio? I need to debug this.

josejulio:
I suppose that call needs to be done when passing the arguments on manageiq-ui-classic.

aljesusg:
But in MiddlewareManager the connect call to default_endpoint.ssl_cert_store that do this transform

josejulio:
From the error mentioned before [1], that function isn't being called. Certainly debug is needed there, sorry for not having more information.

[1] https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100#issuecomment-342205355

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -104,8 +112,12 @@ def self.raw_connect(host, port, username, password, security_protocol, cert_sto
         :verify_ssl     => verify_ssl_mode(security_protocol),
         :ssl_cert_store => cert_store
```
israel-hdez: To fix the SSL error, please change this to:
``` ruby
:ssl_cert_store => if cert_store.kind_of?(String)
                     Endpoint.new(:certificate_authority => cert_store).ssl_cert_store
                   else
                     cert_store
                   end
```

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -51,26 +51,34 @@ class Hawkular::MiddlewareManager < ManageIQ::Providers::MiddlewareManager
...
-      rescue => err
-        $log.error(err)
-        raise MiqException::Error, 'Unable to verify credentials'
-      end
+      validate_connection(connect(options), options[:hostname], options[:port])
 
       true
     end
 
+    def self.validate_connection(connection, hostname = nil, port = nil)
```
israel-hdez: Mark this as `private`

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -51,26 +51,34 @@ class Hawkular::MiddlewareManager < ManageIQ::Providers::MiddlewareManager
...
+      # As the connect will only give a handle
+      # we verify the credentials via an actual operation
+      connection_rescue_block(hostname, port) do
+        connection.inventory.list_feeds
+      end
+    end
+
+    def self.connection_rescue_block(hostname = nil, port = nil)
```
israel-hdez: Mark this as `private`

app/models/manageiq/providers/hawkular/middleware_manager.rb
tumido: Why it should be private? Neither [VmWare](https://github.com/ManageIQ/manageiq-providers-vmware/blob/68f09bf0406e6e4992922157f3d94c0b493eb4c8/app/models/manageiq/providers/vmware/manager_auth_mixin.rb#L57) nor [OpenStack](https://github.com/ManageIQ/manageiq-providers-openstack/blob/bec9f82f8b56b6f2ae537a3d0343101b6984ccc8/app/models/manageiq/providers/openstack/manager_mixin.rb#L40) nor [Amazon](https://github.com/ManageIQ/manageiq-providers-amazon/blob/e46a7cb7f4edc1e9257dfd23789f1f88065c9016/app/models/manageiq/providers/amazon/manager_mixin.rb#L69) nor [Azure](https://github.com/ManageIQ/manageiq-providers-azure/blob/dbb5d39fe0b280b75e7b06d093b11ef6c0d42207/app/models/manageiq/providers/azure/manager_mixin.rb#L39) has these methods as private.

app/models/manageiq/providers/hawkular/middleware_manager.rb
tumido: I think this should come as a separate PR, since it's fixing different issue.

tumido:
@israel-hdez I'll work on the tests, I'm not sure if it should block this to be merged though.

jshaughn:
By applying locally this PR, and also https://github.com/ManageIQ/manageiq-ui-classic/pull/2577, I was finally able to create a non-ssl provider and so hopefully can now make some progress.  It seems it's been a week being unable to create a provider, that is a long time.  If a merge is not imminent I'd suggest maybe posting in dev-list how to work around this problem.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -51,26 +51,34 @@ class Hawkular::MiddlewareManager < ManageIQ::Providers::MiddlewareManager
...
+      self.class.validate_connection(connect(options), options[:hostname], options[:port])
 
       true
     end
 
+    def self.validate_connection(connection, hostname = nil, port = nil)
+      # As the connect will only give a handle
+      # we verify the credentials via an actual operation
+      connection_rescue_block(hostname, port) do
+        connection.inventory.list_feeds
```
jshaughn: @josejulio I'm not sure, but with Inventory V4 is list_feeds available?  This is not a problem for this PR but it may be a problem for applying it to the hawkular-1275 branch.

aljesusg:
Now It's working again after https://github.com/ManageIQ/manageiq-ui-classic/pull/2643 with normal servers and SSL with CA. Tested !!! Do we need this PR?

tumido:
@aljesusg if we want to be prepared for the new dialog, we need this PR. It's not breaking the "old way" it's just enabling the "new way". I vote for being prepared and have your and mine PRs merged.

aljesusg:
Now its working with CA too @tumido

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -94,7 +102,7 @@ def self.entrypoint(host, port, security_protocol)
     end
 
     # Hawkular Client
-    def self.raw_connect(host, port, username, password, security_protocol, cert_store)
+    def self.raw_connect(host, port, username, password, security_protocol, cert_store, validate = false)
       credentials = {
```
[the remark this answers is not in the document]
tumido: This is switched to `true` in the new dialog in your PR, for other calls it should be `false`. So old and new dialogs can live together and it also allows any call to `connect` to handle the connection right.

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -51,26 +51,34 @@ class Hawkular::MiddlewareManager < ManageIQ::Providers::MiddlewareManager
...
     end
 
+    def self.validate_connection(connection, hostname = nil, port = nil)
+      # As the connect will only give a handle
+      # we verify the credentials via an actual operation
+      connection_rescue_block(hostname, port) do
+        connection.inventory.list_feeds
+      end
+    end
+
```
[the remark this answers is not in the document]
tumido: @aljesusg you sure you have the latest revision of this PR?

aljesusg:
Tested with:

https://github.com/ManageIQ/manageiq-ui-classic/pull/2577
https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100
https://github.com/ManageIQ/manageiq-providers-hawkular/pull/101

and revert

https://github.com/ManageIQ/manageiq-ui-classic/pull/2643

and it's working with hawkular server, ssl and CA

app/models/manageiq/providers/hawkular/middleware_manager.rb
tumido: @israel-hdez fixed here https://github.com/ManageIQ/manageiq-providers-hawkular/pull/101

aljesusg:
Tested with:

https://github.com/ManageIQ/manageiq-ui-classic/pull/2577
https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100
https://github.com/ManageIQ/manageiq-providers-hawkular/pull/101

and revert

https://github.com/ManageIQ/manageiq-ui-classic/pull/2643

and it's working with hawkular server, ssl and CA

app/models/manageiq/providers/hawkular/middleware_manager.rb
josejulio: Yes, but i don't think there is other way to test it at this point, "status" is unprotected, so it won't be good way to test the connection.

abonas:
> By applying locally this PR, and also ManageIQ/manageiq-ui-classic#2577, I was finally able to create a non-ssl provider and so hopefully can now make some progress. It seems it's been a week being unable to create a provider, that is a long time. If a merge is not imminent I'd suggest maybe posting in dev-list how to work around this problem.

@jshaughn there was a regression due to a broader change in core. core team fixed it yesterday and merged. so now it should be ok on master. 
in order to fix it to better fix a new dialog behavior for the future, there are several PRs in the works (incl this one).

app/models/manageiq/providers/hawkular/middleware_manager.rb
tumido: @israel-hdez thanks very much for explanation! I didn't realized they are mixins. Cool, I'm making them private now! ;)

tumido: @israel-hdez, ok, I've found out, the "old way" without queued validation is not working when this method is private (because it's called from instance via `verify_credentials`). The other method can be made private for sure.

app/models/manageiq/providers/hawkular/middleware_manager.rb
tumido: fixed

tumido:
@israel-hdez test coverage is available here https://github.com/ManageIQ/manageiq-providers-hawkular/pull/102

app/models/manageiq/providers/hawkular/middleware_manager.rb
cfcosta: This name is not descriptive, as well.

cfcosta: Also, I think it would be a great idea to overwrite `with_provider_connection` with this, so that this error handling is used everywhere.

app/models/manageiq/providers/hawkular/middleware_manager.rb
cfcosta: Private class methods are an anti-pattern. If we want to have this not be part of the interface then we should make it an instance method, or extract connection validation to another class (which I would prefer tbh).

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -94,7 +103,7 @@ def self.entrypoint(host, port, security_protocol)
     end
 
     # Hawkular Client
-    def self.raw_connect(host, port, username, password, security_protocol, cert_store)
+    def self.raw_connect(host, port, username, password, security_protocol, cert_store, validate = false)
```
cfcosta: @israel-hdez do you know if MiQ uses `raw_connect`? If not I think this should be on a different class, probably a delegator. What do you think, @tumido?

If so, it could be changed on a different PR, because it seems to me that adding optional params is increasing cyclomatic complexity on the class that is already our biggest offender on this.

israel-hdez: Yes, they are using it :(

tumido: @cfcosta, basically, the whole new workflow is about shifting the load from UI worker to spawning a queued job (created [here](https://github.com/ManageIQ/manageiq/blob/aa4dab9c75942a2c335d939a5eb112e3d25673df/app/models/mixins/authentication_mixin.rb#L16)) Also it tries to smaller the load by not creating a temporary EMS object for the credentials validation and instead validates them as is. So it requires a class method. And they decided, that using the already available `raw_connect` is the right way. For some providers it was already doing the job without any modifications, for others, like Hawkular, it means these adjustments...

tumido:
Closing, won't merge. Repo abandoned. :no_entry:

PR #101, opened 2017-11-07
Create cert_store if it's not created before validation

aljesusg:
This fix the cert store if it's not created before call the validation method, create the endpoint and assign the certificate_authority

Related with : https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100
Related with: UI https://github.com/ManageIQ/manageiq-ui-classic/pull/2577
Related with: UI https://github.com/ManageIQ/manageiq-ui-classic/pull/2660

cc @tumido , @israel-hdez

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -99,6 +99,11 @@ module ManageIQ::Providers
         :username => username,
         :password => password
       }
+
+      if cert_store.kind_of?(String)
+        cert_store = Endpoint.new(:certificate_authority => cert_store).ssl_cert_store
+      end
+
       options = {
         :tenant         => 'hawkular',
         :verify_ssl     => verify_ssl_mode(security_protocol),


tumido:
@aljesusg Please link the dependent/related PRs in the descriptions.

aljesusg:
@miq-bot assign @abonas

aljesusg:
Tested with:

https://github.com/ManageIQ/manageiq-ui-classic/pull/2577
https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100
https://github.com/ManageIQ/manageiq-providers-hawkular/pull/101

and revert

https://github.com/ManageIQ/manageiq-ui-classic/pull/2643

and it's working with hawkular server, ssl and CA

tumido:
@israel-hdez please review

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -102,7 +102,11 @@ def self.raw_connect(host, port, username, password, security_protocol, cert_sto
       options = {
         :tenant         => 'hawkular',
         :verify_ssl     => verify_ssl_mode(security_protocol),
-        :ssl_cert_store => cert_store
+        :ssl_cert_store => if cert_store.kind_of?(String)
```
tumido: Functional-wise I'm glad the change is that simple, though stylistically I would prefer a bit different way, something like:
```
if cert_store.kind_of?(String)
  cert_store = Endpoint.new(:certificate_authority => cert_store).ssl_cert_store

options = {
  ...
  :ssl_cert_store => cert_store
  ...
}
```

aljesusg: I prefer cert_store.kind_of?(String) ? Endpoint.new(:certificate_authority => cert_store).ssl_cert_store : cert_store ^^ Vote community !!!

aljesusg: Ternary operators before if

aljesusg: @tumido wins...I'll change the code ^^ thanks

abonas:
is this ready for merge? or does it have to be merged with the other related PRs?
should/can we have this in gaprindashvili?

tumido:
@abonas, not yet, there's a change suggestion which was promised to be adjusted by @aljesusg.

tumido:
@abonas, it should be merged together with the other PRs.

The gaprindashvili thing - I'm not sure. Is there a plan to have the new queue validation available for Hawkular in this release @agrare? If yes, it needs to land there, if not it doesn't matter.

agrare:
> Is there a plan to have the new queue validation available for Hawkular in this release @agrare? If yes, it needs to land there, if not it doesn't matter.

Not that I know of but it could always happen.  I'd recommend putting a WIP PR in to ui-classic that adds it to the list of ems_types that we do validate as an umbrella PR that can link to all the dependencies.  That way if we do backport that we make sure to include everything it depends on.

tumido:
@agrare, good idea. Here it is ;) https://github.com/ManageIQ/manageiq-ui-classic/pull/2660

agrare:
:heart: thanks @tumido

app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -99,6 +99,9 @@ def self.raw_connect(host, port, username, password, security_protocol, cert_sto
         :username => username,
         :password => password
       }
+      if cert_store.kind_of?(String)
```
cfcosta: Please add line breaks before and after the block.

cfcosta: Also, is there a way this can be tested? Since it broke something, I think we should have regression tests for it.

cfcosta: Also, what do you think on moving this to `Endpoint` and have it default to `cert_store` if provided? I prefer to have it there, so we could remove conditionals on foreign data.

abonas:
Hi, what's the status of this PR? I see a related PR was merged a few days ago, but here some review changes were requested?

aljesusg:
It was pending from the PR merged yesterday and make this changes but I was thinking about make the move of the function to endpoint or not like @cfcosta requested I am gonna make the changes today and add a test for merge this but I think that is not a good idea to move this to the endpoint because this is only happening to this provider.

abonas:
> It was pending from the PR merged yesterday and make this changes but I was thinking about make the move of the function to endpoint or not like @cfcosta requested I am gonna make the changes today and add a test for merge this but I think that is not a good idea to move this to the endpoint because this is only happening to this provider.

it can wait for tomorrow, and it sounds like there is some more discussion needed if certain things sound like not a good idea to you :)

aljesusg:
I think so, maybe we can merge it after me changes and after we can discuss about move it to Endpoint , What do you think @cfcosta?

tumido:
@aljesusg I agree with @cfcosta, it makes a much more sense to have this handled on the `Endpoint` part and I would suggest to change this PR to do so instead of merging and fixing later... This PR is not intended for gaprindashvili so we have plenty of time to do it right. Also just a reminder this PR belongs under changes in https://github.com/ManageIQ/manageiq-ui-classic/pull/2660 (please add the link into the description).

agrare:
> The gaprindashvili thing - I'm not sure. Is there a plan to have the new queue validation available for Hawkular in this release

Not necessary, and I'd argue stability is more important at this point.

aljesusg:
I updated the description @tumido, ok so I'll do the PR for the Endpoint like @cfcosta requested. Thanks

abonas:
@agrare yes, I agree, I asked the gaprindashvili question a month ago. it has been marked as "no" for quite some time now.

israel-hdez:
> @aljesusg I agree with @cfcosta, it makes a much more sense to have this handled on the Endpoint part and I would suggest to change this PR to do so instead of merging and fixing later...

I don't agree "too much"... Even when I suggested this code to fix the related issue, `Endpoint` is being used as a helper and an instance is created just to re-use `ssl_cert_store` method. Perhaps the code to create certificate stores should be extracted to a module under `lib/` in core-repo. But given that `Endpoint` is part if the core repo, you could try to place a PR and see if the core team accepts.

aljesusg:
Yes...In my experience, they don't merge in core something for a specific provider but we can try it.

app/models/manageiq/providers/hawkular/middleware_manager.rb
aljesusg: @cfcosta linre breaks throw a rubocop error :(

aljesusg:
So @cfcosta, @tumido, @israel-hdez What should I do? I can try to move this to the Endpoint or me can merge this.

app/models/manageiq/providers/hawkular/middleware_manager.rb
israel-hdez: It's not the line break that throws the rubocop errors, it's the whitespaces in the blank line :)

israel-hdez:
@aljesusg Since `Endpoint` is in core repo, my vote is to merge now while you check if core team accepts you PR. If they do, just place another PR here. If not... well, we've already made progress.
Just fix the rubocop warning ;)

app/models/manageiq/providers/hawkular/middleware_manager.rb
aljesusg: I clean the whitespaces and still get the same error ^^

aljesusg: I was cleaning whitespaces and still got error It was confused ^^Fixed !!!

aljesusg:
@israel-hdez is ready now. Thanks

PR #102, opened 2017-11-09
Add test coverage for connect and verify_credentials

tumido:
Enhance test coverage of `MiddlewareManager` to provide test cases for:
- Connection via `:connect` method using non-ssl and ssl verification
- Credentials validation via both `:verify_credentials` and `:raw_connect` covering the exception mapping

Based on suggestion by @israel-hdez [here](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100#pullrequestreview-74590009).
A bit of competition PR to https://github.com/ManageIQ/manageiq-providers-hawkular/pull/86 from @aljesus, replacing 4 test cases with more generic ones.

Depends on https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100

### DIFF
--- a/spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
+++ b/spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
@@ -227,4 +227,107 @@ describe ManageIQ::Providers::Hawkular::MiddlewareManager do
       end
     end
   end
+
+  describe 'connectivity' do
+    let(:ems)      { FactoryGirl.create(:ems_hawkular) }
+    let(:username) { 'abcdefghij' }
+    let(:password) { '1234567890' }
+    let(:host)     { 'awesomehawkular.org' }
+    let(:port)     { 8080 }
+
+    describe '#connect' do
+      let(:endpoint)       { FactoryGirl.create(:endpoint) }
+      let(:http_scheme)    { URI::HTTPS }
+      let(:verify_ssl)     { 1 }
+      let(:ssl_cert_store) { nil }
+      let(:connection_options) do
+        {
+          :entrypoint  => http_scheme.build(:host => host, :port => port).to_s,
+          :credentials => { :username => username, :password => password },
+          :options     => { :tenant => 'hawkular', :verify_ssl => verify_ssl, :ssl_cert_store => ssl_cert_store }
+        }
+      end
+      subject { ems.connect }
+
+      before do
+        allow(ems).to receive_messages(
+          :hostname                => host,
+          :port                    => port,
+          :default_endpoint        => endpoint,
+          :authentication_userid   => username,
+          :authentication_password => password
+        )
+      end
+
+      context 'non-ssl' do
+        let(:http_scheme) { URI::HTTP }
+
+        it { is_expected.to be_connected_to connection_options }
+      end
+
+      context 'ssl-without-validation' do
+        let(:verify_ssl) { 0 }
+        let(:endpoint)   { FactoryGirl.create(:endpoint, :security_protocol => 'ssl-without-validation') }
+
+        it { is_expected.to be_connected_to connection_options }
+      end
+
+      context 'ssl' do
+        let(:endpoint) { FactoryGirl.create(:endpoint, :security_protocol => 'ssl') }
+
+        it { is_expected.to be_connected_to connection_options }
+      end
+    end
+
+    context 'validation' do
+      shared_examples 'validate' do
+        let(:connection) { dup }
+        let(:struct) { dup }
+        before do
+          allow(connection).to receive(:inventory).and_return(struct)
+          allow(ems).to receive(:connect).and_return(connection)
+        end
+
+        context 'failed' do
+          before do
+            allow(struct).to receive(:list_feeds).and_raise(exception)
+          end
+
+          [
+            [::Hawkular::Exception.new("Unauthorized", 401), MiqException::MiqInvalidCredentialsError, /Invalid credentials/],
+            [::Hawkular::Exception.new("Host not found", 404), MiqException::MiqHostError, /Hawkular not found on host/],
+            [::Hawkular::ConnectionException.new(nil), MiqException::MiqUnreachableError, /Unable to connect to*/],
+            [URI::InvalidComponentError, MiqException::MiqHostError, /Host .* is invalid/],
+            [StandardError, MiqException::Error, /Unable to verify credentials/]
+          ].each do |raised, expectation, message|
+            context message do
+              let(:exception) { raised }
+
+              it 'raises exception' do
+                expect { subject }.to raise_error(expectation, message)
+              end
+            end
+          end
+        end
+
+        context 'sucessfull' do
+          before do
+            allow(struct).to receive(:list_feeds).and_return([])
+          end
+
+          it { is_expected.to be_truthy }
+        end
+      end
+
+      describe '#verify_credentials' do
+        subject { ems.verify_credentials }
+        include_examples 'validate'
+      end
+
+      describe '.raw_connect' do
+        subject { described_class.raw_connect(host, port, username, password, nil, nil, true) }
+        include_examples 'validate'
+      end
+    end
+  end
 end
--- a/spec/custom_matchers/be_connected_to.rb
+++ b/spec/custom_matchers/be_connected_to.rb
@@ -1,0 +1,5 @@
+RSpec::Matchers.define :be_connected_to do |connection|
+  match do |client|
+    expect(client.state).to eq connection
+  end
+end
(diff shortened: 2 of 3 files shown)


tumido:
cc @abonas

tumido:
**Note:** the `allow_any_instance_of` is used due to necessity to raise exceptions when the `::Hawkular::Client` is already instantiated. The exception handling happens later in the code so we can't raise errors on `:new`. The way to avoid such `allow_any_instance_of` call would be:
```ruby
let(:client) { ::Hawkular::Client.new(:entrypoint => "", :credentials => {}, :options => {}) }
...
before do
    allow(::Hawkular::Client).to receive(:new).and_return(client)
end
...
it "invalid credentials" do
    allow(client).to receive(:inventory).and_raise(...)
    except...
end
```
If this is prefered by the team, I'm ok to change it. Though I personally prefer to have the `::Hawkular::Client` instantiated by the MiddlewareManager instead of mocking it. This would help in cases when an exception on client becomes risen during the `:new` call. The mocking of whole client may hide it.

tumido:
@miq-bot assign @cfcosta
please review, thanks! ;)

tumido:
@miq-bot add_label test

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -198,4 +198,101 @@ def timeline_domain_expectations(status)
...
+      before do
+        allow(ems).to receive(:hostname).and_return(config[:host])
+        allow(ems).to receive(:port).and_return(config[:port])
+        allow(ems).to receive(:default_endpoint).and_return(endpoint)
+        allow(ems).to receive(:authentication_userid).and_return(config[:username])
+        allow(ems).to receive(:authentication_password).and_return(config[:password])
+      end
+
+      it 'non-ssl' do
+        expect(::Hawkular::Client).to receive(:new) do |params|
```
cfcosta: What do you think on exposing connection status methods on the Hawkular gem? I generally dislike mocking methods, but exposing `connected?` and `connection_params` on the gem would make this test relevant.

We could start with it as a monkey patch and move it back to the gem as soon as we ship InventoryV4.

tumido: If he can use Hawkular client to define circumstances under which the connection is provided, it would make the tests much usable. I like that idea.

tumido:
@cfcosta, the part in your review about mocks - you're referring to the first part only (testing the connection) or also to the second part (for the credentials validation)?

cfcosta:
@tumido for credential validations is a little better, I think a better way to do so is to actually inject the connection class into the method as an optional parameter. Then at least we can remove the dependency on `allow_any_instance_of`.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
israel-hdez: You have the connection parameters as attributes in the client, although in documentation comments they are declared as private:
https://github.com/hawkular/hawkular-client-ruby/blob/a4f5c64242b7f226dba20d25e8d938a49ccc5691/lib/hawkular/base_client.rb#L18-L19

As Cainã suggests, in this case, I think it's better to check the attributes of the returned connection object rather than mocking the `new` method. I guess we could just remove that yardoc private modifier in the client gem, because that data is quite important to be exposed.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -198,4 +198,101 @@ def timeline_domain_expectations(status)
...
+            :credentials => config.slice(:username, :password),
+            :options     => config.slice(:tenant).merge(:verify_ssl => 1, :ssl_cert_store => endpoint.ssl_cert_store)
+          )
+        end
+        ems.connect
+      end
+    end
+
+    context 'validation' do
+      shared_examples 'handles' do |method|
```
israel-hdez: Hmmm... I think we also need a positive test. i.e., a test were the connection succeeds ;)
...and also one test to check that no validation is done if you pass `false` in the last parameter to `raw_connect`.

tumido:
@cfcosta, @israel-hdez, I did a second iteration of this. I hope it's much better, can you please review? Also, it still depends on the https://github.com/ManageIQ/manageiq-providers-hawkular/pull/100 so it will fail for `raw_connect`.

tumido:
@miq-bot add_label gaprindashvili/no

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -227,4 +227,107 @@ def timeline_domain_expectations(status)
...
+      context 'ssl' do
+        let(:endpoint) { FactoryGirl.create(:endpoint, :security_protocol => 'ssl') }
+
+        it { is_expected.to be_connected_to connection_options }
+      end
+    end
+
+    context 'validation' do
+      shared_examples 'validate' do
+        let(:connection) { dup }
```
israel-hdez: Is `dup` creating a double? I think it's the first time I see something like this.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -227,4 +227,107 @@ def timeline_domain_expectations(status)
...
+        end
+      end
+
+      describe '#verify_credentials' do
+        subject { ems.verify_credentials }
+        include_examples 'validate'
+      end
+
+      describe '.raw_connect' do
+        subject { described_class.raw_connect(host, port, username, password, nil, nil, true) }
```
israel-hdez: I looked at my previous comments and I asked for a test where you pass `false` in the last param to check that no validations are done... Not sure if it's a useful test, though.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
tumido: Nope. That's the true power of rspec. 😉  The let statement is lazy evaluated. That means this line would try evaluate the `dup` only once `connection` object is required. But this never happens in our tests. We don't need the `connection` object as is, we only need it to provide us the `inventory` method. And this can be done without being required to define what `dup` really is.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -227,4 +227,107 @@ def timeline_domain_expectations(status)
...
+        {
+          :entrypoint  => http_scheme.build(:host => host, :port => port).to_s,
+          :credentials => { :username => username, :password => password },
+          :options     => { :tenant => 'hawkular', :verify_ssl => verify_ssl, :ssl_cert_store => ssl_cert_store }
+        }
+      end
+      subject { ems.connect }
+
+      before do
+        allow(ems).to receive_messages(
```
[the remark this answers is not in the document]
tumido: Ok, make sense, I'll take look what would be the cost of it. I'm not entirely sure if it's necessary for these tests though.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
```diff
@@ -227,4 +227,107 @@ def timeline_domain_expectations(status)
...
+    let(:ems)      { FactoryGirl.create(:ems_hawkular) }
+    let(:username) { 'abcdefghij' }
+    let(:password) { '1234567890' }
+    let(:host)     { 'awesomehawkular.org' }
+    let(:port)     { 8080 }
+
+    describe '#connect' do
+      let(:endpoint)       { FactoryGirl.create(:endpoint) }
+      let(:http_scheme)    { URI::HTTPS }
+      let(:verify_ssl)     { 1 }
```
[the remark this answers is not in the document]
tumido: I know that this value is verified against these OpenSSL constants, though I'm not sure if that's the real value on input (passed from the UI). I'll check on that.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
tumido: I don't think we need such test since it's not doing any validation with `false`. That would only check if we can create a client instance from given params. Which doesn't make sense to test here.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
israel-hdez: Right. So, basically what you have here are doubles. But it's very confusing :confused: because, indeed, `dup` it's a valid method and it may just run fine and return something.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
israel-hdez: I know, but it's a behavior to cover (code is already covered). How are you ensuring that by passing false, or passing no parameter at all, the validation won't happen?
Although, knowing how this is used, it's really not very relevant. So, it's up to you.

israel-hdez:
To merge this, #100 needs to go in. It it ready? Reading the comments on it, I'm not sure.

spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
cfcosta: If `connection` is defined outside, it will be available here, not sure why the dup.

tumido:
Closing, won't merge. Repo abandoned. :no_entry:

PR #104, opened 2017-11-14
Add .idea to .gitignore

lucasponce:
Trivial, just to clean IDE files based on RubyMine.
If not it can get added in a commit unintentionally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@
 /config/environments/*.local.yml
 
 /spec/manageiq
+
+.idea


cfcosta:
What's our currency policy on this? @abonas 
Imho it should be on each user's global gitignore, since we don't do the same for vim temporary files. Something like this:

.gitconfig:
```
[core]
  excludesfile = ~/.gitignore
```

.gitignore:
```
.idea
```

lucasponce:
https://github.com/ManageIQ/manageiq/blob/master/.gitignore#L10

PR #106, opened 2017-11-15
[1511961] Middleware server reports fixed after introducing STI on MiddlewareServer

xeviknal:
Overriding `supported_models` let us tell to whole MiddlewareServer inheritance that their metrics are listed in the same metrics yaml: [product/live_metrics/middleware_server.yml](https://github.com/ManageIQ/manageiq/blob/master/product/live_metrics/middleware_server.yaml).

Since right now there aren't differences - regarding to live metrics - between EAP/Wildfly there is no need to split file into 3 different ones. When new metrics introduces for only one of the middleware servers kinds, then would be time to split it.

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_server.rb
@@ -17,5 +17,9 @@ module ManageIQ::Providers
         :requesting_user => requesting_user
       )
     end
+
+    def self.supported_models
+      @supported_models ||= ['middleware_server']
+    end
   end
 end


abonas:
please change the title and the commit message to be explanatory that it fixes the problem, right now it sounds like the bug itself ("error ....")

cfcosta:
I don't get it, what is expected to change with this? Can you give it more context?

xeviknal:
sure @cfcosta. When capturing live_metrics for an entity in MIQ we are calling [this](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/live_metrics_mixin.rb#L61) which is calling this yaml parsing [here](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/live_metrics_mixin.rb#L72).

`load_live_metrics_config` is looking a yml file within 'products/live_metrics' folder. Specifically, is looking up for a file named as the instance class name ([here](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/live_metrics_mixin.rb#L58)).

Since in the PR #60 were introduced a STI on MIddlewareServer, now there are two new kids on the neighbourhood: MiddlewareServerWildfly and MiddlewareServerEap. Then, when  `load_live_metrics_config` looks for the yaml name, it finds `middleware_server_wildfly` and `middleware_server_eap` instead of `middleware_server`.  And the files for children in the inheritance are not defined on core repo ([here](https://github.com/ManageIQ/manageiq/tree/master/product/live_metrics)).

israel-hdez:
@xeviknal Can you add the bugzilla link to the main PR description?

simaishi:

```
$ git log -1
commit 83d9f0e5e91581c9ef675ed425b3a1642af99290
Author: Cainã Costa <me@cfcosta.com>
Date:   Wed Nov 15 14:07:37 2017 -0200

    Merge pull request #106 from xeviknal/1510421
    
    [1511961] Middleware server reports fixed after introducing STI on MiddlewareServer
    (cherry picked from commit f16422123da304f70bf4f7ad6893b4ba681995c0)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1514124
```

PR #109, opened 2017-11-22
Add i18n string placeholders test

mzazrivec:
This is just invocation of standard i18n placeholder test, included from main ManageIQ
repository. It's the same thing that we do in ui-classic to test that the provided translations
honor placeholders in the original (English) strings.

### DIFF
--- a/spec/i18n/placeholders_spec.rb
+++ b/spec/i18n/placeholders_spec.rb
@@ -1,0 +1,3 @@
+describe :placeholders do
+  include_examples :placeholders, ManageIQ::Providers::Hawkular::Engine.root.join('locale').to_s
+end


spec/i18n/placeholders_spec.rb
```diff
@@ -0,0 +1,3 @@
+describe :placeholders do
```
cfcosta: `describe 'i18n placeholders'` is better, imho.

abonas:
@mzazrivec can you please explain why is this test needed in this repo? which code/placeholders should it test here? (the repo is server side focused) thanks

mzazrivec:
I don't quite understand what 'server side focused repo' means.

Though what this test (shared spec) does is that it tests that the placeholders in gettext
strings are preserved in translations. That means that for example if you have English string:

`My name is %{name}`

its translation into Spanish has to preserve the placeholder:

`My nombre is %{name}`

In the past, translators would make mistakes during translation: they would skip the placeholder
or translate the string as:

`My nombre is %{nombre}`

or:

`My nombre is {nombre}`.

Either of these mistakes would cause problems in our application once it's switched to non-English
locale.

We have this test in core manageiq as well as ui-classic and it needs to be included in all ManageIQ
plugins (engines).

abonas:
> We already have (run) this test in core manageiq as well as ui-classic and it needs to be included
> in all ManageIQ plugins (engines) that have thei rown gettext catalog.

this is the part that I am not fully understanding - where in this repo there is the gettext catalog?
aren't all translations residing in miq ui classic?

mzazrivec:
There's a shared spec (the test) in core manageiq and there's a gettext catalog in this repo
under `locale/` (every plugin has to have its own catalog). This PR just includes the shared spec,
i.e. this PR causes that shared spec to be run for the gettext catalog in this repo.

karelhala:
@mzazrivec so to understand this correctly, if we were to have some custom texts we should also update our `locale/`.

mzazrivec:
@karelhala If you were to have some custom texts? I'm not quite sure what that means.

This repo contains gettext strings, so it has to have its own gettext catalog. This PR adds
(or well, includes) a spec to test the translated catalog.

If you change the string inside a gettext call: `_("some string...")` or `N_('some string ...')`
you don't have to update the catalog under `locale/`. That update only needs to be done
when you want to upload the catalog to Zanata for translation.

PR #111, opened 2017-11-24
Re-add remove datasource operation

tumido:
Re-add remove datasource operation missing since https://github.com/ManageIQ/manageiq-providers-hawkular/pull/9.

Fixes: [BUG 1515285](https://bugzilla.redhat.com/show_bug.cgi?id=1515285)

I'm not sure though if this is the correct approach, or I should use a separate method in `app/models/manageiq/providers/hawkular/inventory/server_operations.rb` instead. This solution is not firing any error when run multiple times and basically in any case when the result is an error event...

@cfcosta, please review

### DIFF
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -48,6 +48,8 @@ module ManageIQ::Providers
     group_operation :suspend, 'Suspend Servers'
     group_operation :resume, 'Resume Servers'
 
+    specific_operation :remove_middleware_datasource, 'RemoveDatasource'
+
     attr_accessor :client
 
     def verify_credentials(_auth_type = nil, options = {})


tumido:
@miq-bot add_label bug, operations

cfcosta:
LGTM, seems like I missed it. Thanks!

tumido:
@miq-bot add_label gaprindashvili/yes

@cfcosta, should I somehow address the operation result? Because right now it's successful in MIQ no matter what's the result of the operation in Hawkular, nor notification is posted (which I would expect).

image

cfcosta:
@tumido AFAIK, the operation flash is always successful because it signals that the operation was sent, not that it was complete (this happens with a Notification). Can you check to see if the old implementation behaves the same? If not, then disregard everything I said and yeah, it should address the operation result.

If it does, however, we might need to change the text or see with @abonas what's the desired outcome.

abonas:
@cfcosta @tumido I think that the notification should state that the request was successfully *sent*, because as you rightfully say above, we don't know at this point that it was actually *done* successfully.
I think we should change it to that even if in the past it was saying something different that perhaps was misleading to our users.
@serenamarie125 - ack?

tumido:
@abonas, yes, I totally agree with that. This PR should cover it: https://github.com/ManageIQ/manageiq-ui-classic/pull/2832

@cfcosta, I meant that toast is posted, as you can see on the screenshot, but no Notification at all due to the following: It's because the `specific_operation` is run on `MiddlewareServer`, but this operation in invoked from `MiddlewareDatasource` controler. So it silently failed to get proper MW server instance ([here](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/73e47ee2cba5abd8a5b966f80e50467e91f6ceb5/app/models/manageiq/providers/hawkular/inventory/operation_notification.rb#L16)) and no notification is posted. Also before #9 it indeed behaved the same way as it is with this PR. This PR aims to solve it: #112

simaishi:

```
$ git log -1
commit 0f4e287a9120e79d1629fbdbaa06358275765e89
Author: Cainã Costa <me@cfcosta.com>
Date:   Fri Nov 24 16:33:59 2017 -0200

    Merge pull request #111 from tumido/remove_datasource
    
    Re-add remove datasource operation
    (cherry picked from commit 73e47ee2cba5abd8a5b966f80e50467e91f6ceb5)
```

PR #112, opened 2017-11-24
Fix RemoveDatasource operation notification

tumido:
This PR rewrites the MW datasource removal operation, so it is run against MW datasource (because it's called from MW datasource controller, not the MW server one). It ensures proper `Notification` is posted, and also fixes the notification handling for MW deployment (because solution to this problem fixes that as well)

Additional fix enhancement to #111's issue [~~BUG 1515285~~](https://bugzilla.redhat.com/show_bug.cgi?id=1515285)
Fixes [BUG 1518292](https://bugzilla.redhat.com/show_bug.cgi?id=1518292)

image

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
@@ -12,7 +12,7 @@ module ManageIQ
 
           def emit
             ActiveRecord::Base.connection_pool.with_connection do
-              mw_entity = args.entity_klass.find_by(:id => args.target_resource) unless args.entity_klass == MiddlewareServer
+              mw_entity = args.entity_klass.find_by(:ems_ref => args.target_resource) unless args.entity_klass == MiddlewareServer
               mw_server = if mw_entity.nil?
                             MiddlewareServer.find_by(:ems_ref => args.target_resource) ||
                               MiddlewareDomain.find_by(:ems_ref => args.target_resource)
--- a/app/models/manageiq/providers/hawkular/inventory/server_operations.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/server_operations.rb
@@ -24,10 +24,10 @@ module ManageIQ
               end
             end
 
-            def specific_operation(name, action_name, default_params = {})
+            def specific_operation(name, action_name, default_params = {}, extra_data = {})
               define_method(name) do |ref, params = {}|
                 params[:resourcePath] = ref.to_s
-                run_operation(default_params.merge(params), action_name)
+                run_operation(default_params.merge(params), action_name, extra_data)
               end
             end
           end
@@ -253,10 +253,10 @@ module ManageIQ
           def run_operation(parameters, operation_name = nil, extra_data = {})
             with_provider_connection do |connection|
               notification_args = NotificationArgs.success(
-                extra_data[:original_operation] || parameters[:operationName],
+                extra_data[:original_operation] || parameters[:operationName] || operation_name.try(:titleize),
                 nil,
                 extra_data[:original_resource_path] || parameters[:resourcePath],
-                MiddlewareServer
+                extra_data[:original_klass] || MiddlewareServer
               )
 
               operation_connection = connection.operations(true)
--- a/app/models/manageiq/providers/hawkular/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager.rb
@@ -48,7 +48,7 @@ module ManageIQ::Providers
     group_operation :suspend, 'Suspend Servers'
     group_operation :resume, 'Resume Servers'
 
-    specific_operation :remove_middleware_datasource, 'RemoveDatasource'
+    specific_operation :remove_middleware_datasource, 'RemoveDatasource', {}, :original_klass => MiddlewareDatasource
 
     attr_accessor :client
 
--- a/spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
+++ b/spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb
@@ -68,6 +68,14 @@ describe ManageIQ::Providers::Hawkular::MiddlewareManager do
         :ems_ref               => '/t;hawkular/f;master.Unnamed%20Domain/r;Local~~/r;Local~%2Fhost%3Dmaster'
       )
     end
+    let(:mw_datasource) do
+      FactoryGirl.create(
+        :hawkular_middleware_datasource,
+        :ext_management_system => ems,
+        :server_id             => mw_server.id,
+        :ems_ref               => '/t;hawkular/f;master.Unnamed%20Domain/r;Local~~/r;Local~%2fsubsystem%3ddatasources%2fdata-source%3dExampleDS'
+      )
+    end
 
     before(:all) do
       MiqServer.seed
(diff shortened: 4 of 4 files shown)


app/models/manageiq/providers/hawkular/middleware_manager.rb
```diff
@@ -48,8 +48,6 @@ class Hawkular::MiddlewareManager < ManageIQ::Providers::MiddlewareManager
     group_operation :suspend, 'Suspend Servers'
     group_operation :resume, 'Resume Servers'
 
-    specific_operation :remove_middleware_datasource, 'RemoveDatasource'
-
```
abonas: there was a special handling for remove? but not for other operations?

app/models/manageiq/providers/hawkular/inventory/server_operations.rb
```diff
@@ -27,6 +27,7 @@ def generic_operation(name, action_name, default_params = {}, default_extra_data
             def specific_operation(name, action_name, default_params = {})
               define_method(name) do |ref, params = {}|
                 params[:resourcePath] = ref.to_s
+                params[:operationName] = action_name
```
abonas: how did this work for other operations before you added this line? it looks like a generic place for not only the remove operation.
or it did not work for other ops as well?

abonas:
is there any unitest that can "catch" this regression shall it happen again? something that validates that the remove operation exists and is supported? (without the call to real hawkular server, etc)

abonas:
iirc you need to link the commit msg to the BZ that it fixes so they can be linked properly

abonas:
@tumido you probably need to add here the screenshot/elaborate what notification is this PR about :)

tumido:
@abonas, it's still WIP, I'll include everything screenshots, description, etc., once I figure out if I'm not introducing some regression with ems_ref passing. I also need to investigate on the tests. Expect it to be ready for review by tommorow. ;)

app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
```diff
@@ -12,7 +12,7 @@ def initialize(args, manager)
 
           def emit
             ActiveRecord::Base.connection_pool.with_connection do
-              mw_entity = args.entity_klass.find_by(:id => args.target_resource) unless args.entity_klass == MiddlewareServer
+              mw_entity = args.entity_klass.find_by(:ems_ref => args.target_resource) unless args.entity_klass == MiddlewareServer
```
tumido: This line is actually fixing notification parsing for all cases when the `ems_ref` was not `MiddlewareServer` (all operations which does not go via MiddlewareServer controler), such as:
- remove MW datasource
- undeploy, disable, enable, restart of MW deployment

app/models/manageiq/providers/hawkular/middleware_manager.rb
tumido: yes, it was introduced by me in #111, but it's wrong because the remove action can't be run on MW server because the controller is different (the operations works fine but the notifications are not)

app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
tumido: @abonas, do we have a ticket for that ^ as well? Because without this change no notifications are posted for MW deployment operations neither...

tumido:
@abonas, also about the test coverage: we don't have any specs for [server_operations.rb](https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/server_operations.rb) at all. That means we're not verifying neither our operations presence, nor if they map to correct hawkular-client-ruby, nor if notifications and event emitter is called with proper data... That's why we didn't noticed this issue beforehand.

cfcosta:
Given that removing middleware datasource is the only specific operation implemented, can this be generalized so that `specific_operation` implements this?

tumido:
@cfcosta, sure it can! ;) Here you go!

app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
abonas: @tumido not afaik. if it's needed, please open a bz so qe can understand the full context of the fix that applies to other entities too.

tumido: @abonas, I hope I did it right: [BUG 1518292](https://bugzilla.redhat.com/show_bug.cgi?id=1518292)

israel-hdez: Perhaps, could you add a small test to check that the notification is placed? (At least, for the _remove datasource_ operation that you are fixing)

I guess it may look like these ones:
https://github.com/ManageIQ/manageiq-providers-hawkular/blob/f02dd94d3b248878d18d5a2287fa0a1b56373414/spec/models/manageiq/providers/hawkular/middleware_manager_spec.rb#L176-L186

cfcosta:
@israel-hdez you bring up a good point, I think having tests for this is necessary, because the first bug happened because the existing tests did not cover this method.

@tumido what do you think? I think that it would add enough value to be worth adding.

tumido:
@israel-hdez, @cfcosta I definitely think a test suite for operations is needed. And since it seems there's more test coverage missing than just this method, I'm willing to write it. But I don' think we should block this PR on that. Once I'm done with bugs, I'll finish the test coverage for connection #102 and follow with this one? What do you think about that?

@abonas, should I track the test tasks somewhere? In JIRA Hawkular project I assume, correct?

abonas:
yes I also [raised concerns ](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/112#issuecomment-346997727)about lack of tests.
What I suggest doing is adding a test specific for this use case either as part of this PR or as an immediate followup. And then, definitely another more extensive task (after other tasks mentioned above) to add testing for other operations.

cfcosta:
@tumido not for all operations, I've been thinking about only expanding the current test suite to include a test for this operation, most of the other ones do have one already.

tumido:
OK :) I'll add a testcase for this operation (as a part of this PR - tomorrow, today's the testathon, yay!) and later we can elaborate on better test coverage for everything! ;)

abonas:
@committers-hawkular I'm testing the maintainers team notification, please ignore :)

tumido:
@israel-hdez, should I be fixing this? It's the same syntax we use on lines [36](https://github.com/tumido/manageiq-providers-hawkular/blob/f7bf87030909c1af48edf969f8852210ceb0ee17/app/models/manageiq/providers/hawkular/middleware_manager.rb#L36), [37](https://github.com/tumido/manageiq-providers-hawkular/blob/f7bf87030909c1af48edf969f8852210ceb0ee17/app/models/manageiq/providers/hawkular/middleware_manager.rb#L37).

israel-hdez:
@tumido Oh! We either, missed the [rubocop comments in the original PR](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/9#issuecomment-338319822), or just ignored them for the sake of syntax consistency with rails methods. So... Not sure. Duh!

WDYT, @cfcosta @abonas ?

abonas:
@tumido the original BZ is already closed and verified. this PR is marked gaprindashvili/yes and bug. is there another bug that is tracked for this specific fix that is also targeted at the gaprindashvili?

tumido:
@abonas, sure, updating the description, the BZ is [1518292](https://bugzilla.redhat.com/show_bug.cgi?id=1518292) ;)

simaishi:

```
$ git log -1
commit d1c693efd64b66f6083fc9ea64d997310256bf1d
Author: Edgar Hernández <ehernand@redhat.com>
Date:   Wed Dec 6 23:45:01 2017 -0600

    Merge pull request #112 from tumido/remove_datasource
    
    Fix RemoveDatasource operation notification
    (cherry picked from commit 888a77c9a2013bc9a25079da831e19afba22eb00)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1524625
```

PR #117, opened 2017-12-05
[WIP] Indicate of MiddlewareServer of a Event

aljesusg:
This PR will change to try fix the link

Depends of UI [#2958](https://github.com/ManageIQ/manageiq-ui-classic/pull/2958)
We need to add more information to the events about servers (such as hostname, or make some events clickable if server already exists in the inventory).

JIRA [JMAN4-242](https://issues.jboss.org/browse/JMAN4-242)
middleware

cc @abonas

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
@@ -33,13 +33,15 @@ module ManageIQ
               unless mw_entity
                 EmsEvent.add_queue(
                   'add', manager.id,
-                  :ems_id          => manager.id,
-                  :source          => 'HAWKULAR',
-                  :timestamp       => Time.zone.now,
-                  :event_type      => args.event_type(mw_server),
-                  :message         => args.event_message(mw_server),
-                  :middleware_ref  => mw_server.ems_ref,
-                  :middleware_type => mw_server.class.name.demodulize
+                  :ems_id                 => manager.id,
+                  :source                 => 'HAWKULAR',
+                  :timestamp              => Time.zone.now,
+                  :event_type             => args.event_type(mw_server),
+                  :message                => args.event_message(mw_server),
+                  :middleware_server_id   => mw_server.id,
+                  :middleware_server_name => mw_server.name,
+                  :middleware_ref         => mw_server.ems_ref,
+                  :middleware_type        => mw_server.class.name.demodulize
                 )
               end
             end
--- a/app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
+++ b/app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
@@ -88,14 +88,16 @@ module ManageIQ::Providers
 
       EmsEvent.add_queue(
         'add', ems_id,
-        :ems_id          => ems_id,
-        :source          => 'EVM',
-        :timestamp       => Time.zone.now,
-        :event_type      => 'hawkular_event',
-        :message         => _('Generation of JDR report was requested by a user.'),
-        :middleware_ref  => middleware_server.ems_ref,
-        :middleware_type => 'MiddlewareServer',
-        :username        => requesting_user
+        :ems_id                 => ems_id,
+        :source                 => 'EVM',
+        :timestamp              => Time.zone.now,
+        :event_type             => 'hawkular_event',
+        :message                => _('Generation of JDR report was requested by a user.'),
+        :middleware_server_id   => middleware_server.id,
+        :middleware_server_name => middleware_server.name,
+        :middleware_ref         => middleware_server.ems_ref,
+        :middleware_type        => 'MiddlewareServer',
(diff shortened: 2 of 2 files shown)


aljesusg:
With this is working in all cases unless alerts. We need the ids of Middleware Server, maybe we can get them with https://github.com/ManageIQ/manageiq-providers-hawkular/pull/103. I am thinking about.

aljesusg:
@miq-bot remove_label wip

abonas:
@miq-bot add_label gaprindashvili/no

app/models/manageiq/providers/hawkular/middleware_manager/middleware_diagnostic_report.rb
```diff
@@ -88,14 +88,16 @@ def enqueue_job
...
-        :middleware_ref  => middleware_server.ems_ref,
-        :middleware_type => 'MiddlewareServer',
-        :username        => requesting_user
+        :ems_id                 => ems_id,
+        :source                 => 'EVM',
+        :timestamp              => Time.zone.now,
+        :event_type             => 'hawkular_event',
+        :message                => _('Generation of JDR report was requested by a user.'),
+        :middleware_server_id   => middleware_server.id,
+        :middleware_server_name => middleware_server.name,
```
abonas: is this the only really changed line of code? the rest seems like formatting changes. if so, it's confusing, please leave only the "real" changed code

app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
```diff
@@ -33,13 +33,15 @@ def emit
...
-                  :message         => args.event_message(mw_server),
-                  :middleware_ref  => mw_server.ems_ref,
-                  :middleware_type => mw_server.class.name.demodulize
+                  :ems_id                 => manager.id,
+                  :source                 => 'HAWKULAR',
+                  :timestamp              => Time.zone.now,
+                  :event_type             => args.event_type(mw_server),
+                  :message                => args.event_message(mw_server),
+                  :middleware_server_id   => mw_server.id,
+                  :middleware_server_name => mw_server.name,
```
abonas: wasn't the original request for the hostname  to be added? (which is not the server name, it's another property), or it's converted somehow on the way?

aljesusg: I asked Jdoyle in Kira about because I thought it was the hostname (I can do it too without link to the host inventory) but it's make sense to be the middleware server.

israel-hdez: These two fields, `middleware_server_id` and `middleware_server_name` were previously being set in [EmsEvent class](https://github.com/ManageIQ/manageiq/blob/master/app/models/ems_event.rb#L146-L155). Indeed, I remember the server was correctly being linked in the events.
Perhaps, this is an issue injected by #60 and may be better to fix `EmsEvent` for a broader scope.

aljesusg: Mmm after https://github.com/ManageIQ/manageiq-ui-classic/pull/2958 we can see the middleware server in user operations, maybe this is different @israel-hdez

israel-hdez:
BTW, this PR is dealing only with the operations events. But the JIRA also refers to the events coming from the event catcher.

app/models/manageiq/providers/hawkular/inventory/operation_notification.rb
israel-hdez: ManageIQ/manageiq-ui-classic#2958 is doing the same: it's setting manually those fields. So, if you fix `EmsEvent` you'll be fixing both the UI part and the events in this PR... And any other similar events.

israel-hdez: (Try reverting #60 in your env and, most likely, you will see the links to the server without the need of these PRs)

aljesusg: The only thing I see is  inventory_object.assign_attributes maybe this is the different, ok I am gonna check this. thanks

aljesusg:
Which events ? The alerts? for this we need to make more changes in alert_manager , in JIRA says "The bug above deals with "new server added" event, but other events like power operations request or power operations result and others should also have full information about which server is it related to." so I only did the changes for all kind of operations but I think will be great to have this for alerts too.

israel-hdez:
There are some events that come from the event catcher without the need to configure an alert in MiQ. The aavailability changed events are one example. Though, not sure if there are others.

aljesusg:
@miq-bot add_label wip

aljesusg:
I did 

https://github.com/ManageIQ/manageiq-ui-classic/pull/3093
https://github.com/ManageIQ/manageiq-providers-hawkular/pull/123

to solve the Jira request, I am gonna focus not to try to fix the EmsEvent like @israel-hdez  said

aljesusg:
Closed by https://github.com/ManageIQ/manageiq/pull/16676

PR #118, opened 2017-12-11
[WIP] Generic UI: GVM and Persistence

cfcosta:
This is a mega-PR, including the whole of Generic View Mapping for the Provider code. The idea is, as soon as this is finished, to be a drop-in replacement for the current inventory parser/collector to persist the inventory using GVM, with no difference on the backend, which means that the runtime works differently, but persists the exact same thing.

This includes #74 as a baseline to get GVM working, but expands on that.

Current Status:

* [x] Integrate GVM to fetch and show data (this is the part where Mike's code is built upon).
* [x] Rewrite parser to use the defined entities.
* [x] Iron out errors and wrong behaviours (manual testing).
* [ ] Add proper testing for new code.
* [ ] Rewrite git history to make things make sense.
* [ ] Remove old workflow tests
* [ ] Merge!

Known-bugs and concerns:

* **Why does this needs new tests? Why isn't the old tests good enough for this?**
  One change on the persistence behaviour is that all requests to the server are done upfront, so fetching all resources, config and so on is done only once, and those will generate GVM entities. GVM entities are normal ruby objects, without being tied to Hawkular or the database. That means that, to test inventory, we only need to test hawkular stuff once, on `HawkularConnection`. Every other part of the inventory persistence talks only in terms of POROS, and those can be created for isolated testing.

  That also means that for most of the inventory refresh code, VCR will not be required.

* **There is some hardcoding of entity types, specially in relation to Domains, why?**
  This is expected, as doing so by being totally generic is not possible yet, at least not without changing the flow of the sync, which I decided not to do yet. After the new testing suite is implemented for this flow, we can refactor it and remove all non-generic parts with confidence.

* **Why is 0d2a39d here, shouldn't it be removed?**
  It will be before merging this PR, it's only being done so to be able to test the gem without Rails/EVM.


cfcosta:
Update:

Finished manual testing, the inventory is working as expected. The data is the same as before on all resources.

Before:


After:

Ladas:
@cfcosta just FYI, this will probably change in the future. The goal for 5.0 is to separate to provider specific collector and generic core persistor workers. Then the collector will be sending data for saving in small batches. The process will be called Streaming Refresh, when neither collector nor persitor will hold the full inventory. 

So "all requests to the server are done upfront" will not be viable, since using streaming refresh, we want to get to a flat memory consumption. So collector should use paginated/streamed api call, parse it, send it to queue and forget it.

This is needed thing for bigger envs, because right now, both collector and and persistors are rising in memory with the size of envs. (we've seen 14GB RAM worker and we expect bigger envs) 

------

So far we had just a Poc, where the process looked like 
1. add data to InventoryCollection of Persistor
2. once Persitor has X elements, call Peristor.to_json and send it to Persistor queue
3. Init new Peristor with InventoryCollections
4. jump to step 1

And the collector will need to work the similar way, with reading only 1 batch of data from the api and keeping only this batch in memory.

cfcosta:
@Ladas oh, I see, this does make sense, and I didn't know the exact plans but they're still aligned with what I did here. The parser methods work on a list of servers/feeds, but those might not be the complete ones. So for a streaming solution, that'd require small changes, just some middle-man between ResourceCollection and the parser, to fill the queue to be processed.

Right now I don't think there's a way to paginate/stream resources from Hawkular, but I'll definitely keep this in mind.

lib/manageiq/providers/hawkular/entities/wildfly_server.rb
```diff
@@ -0,0 +1,26 @@
...
+    module Hawkular
+      module Entities
+        # A specialized version of MiddlewareServer for servers running WildFly.
+        class WildflyServer < MiddlewareServer
+          # Returns if this entity is appliable for given metadata.
+          def self.applicable?(data)
+            data[:type_path].to_s.end_with?(';WildFly Server')
+          end
+
+          def immutable?
```
israel-hdez: Well... This explains why you are using `immutable?` as an alias for `in_container`. But it's not the same :)

lib/manageiq/providers/hawkular/entity_mapper.rb
```diff
@@ -0,0 +1,37 @@
...
+        def self.map(data)
+          new(data).map
+        end
+
+        def initialize(data)
+          @data = data.dup.symbolize_keys
+        end
+
+        def map
+          feed = @data[:resourceTypePath].match(/;([0-9a-zA-Z.\s%]+)\/rt/)[1]
```
israel-hdez: I think the right regex for the feed is `/f;([^/])/` (didn't tested it). But I think you should use `CanonicalPath` class of the ruby gem as a helper to extract the feed.

Anyway, this will need to change, because for inventory v4 there are no canonical paths and the feed cannot be inferred from the paths/ids.

lib/manageiq/providers/hawkular/resource_collection.rb
```diff
@@ -0,0 +1,40 @@
...
+            GenericViewMapper.matcher.find_entity_for(data).new(data)
+          end
+        end
+
+        def prepared
+          servers = entities.select { |x| x.kind_of?(Entities::MiddlewareServer) }
+          resources = (entities - servers).group_by(&:feed)
+
+          servers.each do |server|
+            server.relationships = resources[server.feed]
```
israel-hdez: Remember that the feeds aren't the servers.
For the domain case, most likely all domain servers are being assigned the same set of resources.

lib/manageiq/providers/hawkular/resource_collection.rb
```diff
@@ -0,0 +1,40 @@
...
+
+            resources[server.feed]
+              .inject({}) { |accum, el| accum.merge(el.properties) }
+              .merge!(server.properties)
```
israel-hdez: What's this doing? I don't see where the accumulated properties hash ends, so I think it just gets collected by the garbage collector.

israel-hdez:
I think I need to see how is coded the UI side. Just by looking the provider side code, it's very hard for me to understand why the refresher needs to change. This is adding one more processing layer to the refresher and I think that's not very desirable. I was expecting to see the GV mappings around the MiQ database entities rather than the Hawkular entities (which are external to MiQ and "converted" to MiQ entities in the parsing process).

Also, I agree with @Ladas that it's not very viable to do all server requests in advance. It also has the inconvenient that it'll be hard to adapt my targeted refresh work to this work.

Anyway, I think you'll need to rebase this work on top of Hawkular's Inventory v4 work. It's becoming ready to land in the master branch.

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -1,402 +1,185 @@
...
-        parse_base_item(messaging, inventory_object)
-        inventory_object.name = messaging.name
-
-        type_path = ::Hawkular::Inventory::CanonicalPath.parse(messaging.type_path)
-        inventory_object.messaging_type = URI.decode(type_path.resource_type_id)
-
-        if !config.empty? && !config['value'].empty? && config['value'].respond_to?(:except)
-          inventory_object.properties = config['value'].except('Username', 'Password')
+      def find_updated_resource(resource_name)
+        collector.send("#{resource_name}_updates").each do |item|
```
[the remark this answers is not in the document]
cfcosta: Yeah, I did revert those.

lib/manageiq/providers/hawkular/connection.rb
```diff
@@ -0,0 +1,51 @@
...
+        include ::Hawkular::ClientUtils
+
+        attr_reader :client
+
+        def initialize(client = connect)
+          @client = client
+        end
+
+        def connect
+          ::Hawkular::Client.new(
```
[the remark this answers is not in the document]
cfcosta: Yeah, that is going to be removed when I finish implementing it.

lib/manageiq/providers/hawkular/entities/wildfly_server.rb
cfcosta: That's true! Will fix.

lib/manageiq/providers/hawkular/entity_mapper.rb
cfcosta: There's a second branch that I did implement InventoryV4 for GVM. With it, EntityMapper is going to be gone.

https://github.com/cfcosta/manageiq-providers-hawkular/pull/2

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -1,402 +1,185 @@
...
-        [
-          machine_id[0, 8],
-          machine_id[8, 4],
-          machine_id[12, 4],
-          machine_id[16, 4],
-          machine_id[20, 12]
-        ].join('-')
-      end
+      def fetch_domains
+        collector.all_resources.each do |server|
```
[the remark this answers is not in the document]
cfcosta: This is one thing that doesn’t make sense for me, does the current implementation does this? From what I saw, it only inventories domain_servers, so I might be missing something.

lib/manageiq/providers/hawkular/connection.rb
```diff
@@ -0,0 +1,51 @@
...
+            :options     => { :tenant => 'hawkular' }
+          )
+        end
+
+        def fetch_server(feed)
+          fetch_all_resources([feed]).first
+        end
+
+        def fetch_all_resources(feeds = client.inventory.list_feeds)
+          feeds.flat_map do |feed|
```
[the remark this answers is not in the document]
cfcosta: This is on purpose. Given that we’re only generating normal ruby objects, both the heap size and allocation count is not that big, specially in comparison with old sync.

lib/manageiq/providers/hawkular/resource_collection.rb
cfcosta: This is done for view-related purposes, to cover cases such as when we need to show if a server is immutable or not, since it’s not being defined on the Server entity, but on the OperatingSystem resources.

cfcosta:
@israel-hdez Thanks for the review!

So, for the UI part of the code, you can check @mtho11 ’s pull request at https://github.com/ManageIQ/manageiq-ui-classic/pull/2702

The idea for this PR is to use the same set of intermediate representation that we are using on the view layer to persist the data into the database. It is a first step on having GVM entities to be the de-facto representation for everything inventory.

I did this PR first to guarantee that we can represent the current server db schema using only GVM. The nexe steps is to adapt the DB schema to use a representation that’s compatible with GVM the other way around, generating the object tree from the database.

PR #121, opened 2017-12-13
[WIP] Alert sync refresher

xeviknal:
[WIP] Description
Added a PostRefresh hook in charge of sync'ing alerts and alert profiles for a ems with hawkular backend.

This PR fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1510421
https://bugzilla.redhat.com/show_bug.cgi?id=1515396

Scenarios meant to cover by the PR:
 [Predefined HAlerts](https://docs.google.com/presentation/d/1hch3Qf9vTeU3CBGvtqUOlloIproAKejxKIruSM2dcWA/edit#slide=id.g2aeadbbf72_0_5 ) (thanks to @israel-hdez)

Depending on:
ManageIQ/manageiq#16662

@miq-bot add_label alerts, wip, gaprindashvili/no


israel-hdez:
@jshaughn @lucasponce @josejulio 
breath deeply and... quickly look at this

josejulio:
i think this work should be based on hawkular-1275 branch, is not intended for gapri.... release. Right?

jshaughn:
> i think this work should be based on hawkular-1275 branch, is not intended for gapri.... release. Right?

Yes, I confirmed this with @abonas.   So, in this PR @xeviknal did a lot of refactoring of the old, Hawkular Metrics based, trigger definitions.   The refactoring is nice and can mostly be applied to the new, Prometheus based, definitions.  So that is good.  But, I can't recommend that we actually merge this to master.  The Hawkular-1328 (GA/Prometheus) work is reaching a merge point itself.   I think any work being done for GA on the provider team should be working against the 1328 branches and not master.  I realize this work probably started when the BZs were targeted for TP3  but that is no longer true.  If we merge this to master it will never actually see a release.  Instead, it will just generate massive conflicts when the 1328 work is merged to master for the GA work.

I'll leave the actual ruby-based review to the ruby guys.  It would be helpful though, @xeviknal, if the PR (or BZs, or some doc) provided a high-level description of the overall change in mechanism.  My only comment here is that we try and minimize work on each refresh.  It may already be true but we definitely don't want to update triggers that have already been defined.   And we should avoid fetching the whole trigger population.  In short, we want to work with the resource delta between refreshes.

lucasponce:
I think this PR should be adapted and merge on top of the HAWKULAR-1328 as well.

xeviknal:
> The refactoring is nice and can mostly be applied to the new, Prometheus based, definitions. So that is good.

Cool, thanks. The usage of builders will make it easier to merge with the new Prometheus based definitions.

>  If we merge this to master it will never actually see a release. Instead, it will just generate massive conflicts when the 1328 work is merged to master for the GA work.

Totally agree on this.

> I think this PR should be adapted and merge on top of the HAWKULAR-1328 as well.

I'll make a PR on top of israel-hdez/manageiq-providers-hawkular#hawkular-1328

> It would be helpful though, @xeviknal, if the PR (or BZs, or some doc) provided a high-level description of the overall change in mechanism.

Work in progress as well :+1: 

> In short, we want to work with the resource delta between refreshes.

I see, now the sync is very overkill. In order to do small updates without fetching the whole triggers population, we should keep on MiQ side which alerts, alerts profiles should be sync'ed every time there is a change (i.e. update alert, remove assignment from alert profile).

lucasponce:
> I see, now the sync is very overkill. In order to do small updates without fetching the whole triggers population, we should keep on MiQ side which alerts, alerts profiles should be sync'ed every time there is a change (i.e. update alert, remove assignment from alert profile).

Problem here will be same as inventory. What happens in something is corrupted in Hawkular ? (or deleted).
So, deltas are good in general but to cover those cases, I don't see any way but sync all MiQ in Hawkular.

xeviknal:
Thinking of covering situations like assigning alerts to alert profiles, assigning resources to alert profiles, updating alerts and alert profiles.

Do you think it could be a good idea to have a sync? 
For example, if you change something of an alert profile, sync the whole profile in cascade (each alert for each resource attached - it might involve update different hawkular providers).
In case you update an alert, sync all the alert profiles which it is assigned to (given that alert, navigate to each alert profile and to a Alert Profile sync).
Using an approach like that, you avoid to sync whole MIQ alerts into Hawkular. You are narrowing down the sync scope but sync'ing in a robust way, making sure that the whole profile will work.
However, full alert sync will have to be performed when a new provider is added (in order to cover those profiles assigned to Enterprise).

In addition, trigger a provider refresh after performing an action to an alert or alert profile, it is not that trivial. Because an alert might be assigned to multiple providers. Also with alert profiles: they might be assigned to servers from different providers. Therefore, we would have situation where to trigger as many provider refreshes as different providers.

Thus, my proposal is to implement a background jobs for:
- AlertProfileSync: syncs assigned alerts to whole assigned resources (it might be in different providers)
- AlertSync: triggers a AlertProfileSync for each profile which holds the alert itself.

Do you guys think that would cover as well corruption in hawkular side and at the same time we are narrowing down the sync scope?

israel-hdez:
@xeviknal 
Not sure why the profile needs to be synced when an alert is updated. I haven't reviewed your code, so I'll comment based on the decisions of the last meeting about this.

Current behavior is that when an alert is created/updated/deleted, the same action is forwarded to Hawkular - to all providers registered in MiQ. No need to go to the profile, nor no need a profile to exist. So, if you include this to a post-refresh process, you need to ensure that all alerts have a corresponding group trigger in all registered Hawkulars, and clean those that aren't needed anymore.

The profile are only relationships between the alerts and the middleware entities. To sync the profile, the idea is to ensurrefreshe that each Hawkular group trigger has the same relationships/members defined in the MiQ profile. 

I believe it isn't too complicated if a full sync is done. 

@lucasponce 
> Problem here will be same as inventory. What happens in something is corrupted in Hawkular ? (or deleted).
So, deltas are good in general but to cover those cases, I don't see any way but sync all MiQ in Hawkular.

IIRC, this was the reason why we decided to move the alert syncing to the provider refresh. Also, MiQ doesn't have enough data to create a delta. It still needs to fetch data from Hawkular to resolve a delta and if we need to fetch data... well, a full refresh makes more sense.

xeviknal:
Closing the branch! thanks for your time :clap:

PR #122, opened 2017-12-18
Added Use case for suspended state in Middleware Servers

aljesusg:
We can do this instead https://github.com/ManageIQ/manageiq-ui-classic/pull/2996

mw_suspended

cc @israel-hdez @tumido  Can you check if this is ok?

### DIFF
--- a/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
+++ b/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
@@ -194,7 +194,7 @@ module ManageIQ::Providers
           props = server.properties
 
           props['Availability'], props['Calculated Server State'] =
-            process_server_availability(props['Server State'], availability.try(:[], 'data').try(:first))
+            process_server_availability(props['Suspend State'] == 'SUSPENDED' ? 'suspended' : props['Server State'], availability.try(:[], 'data').try(:first))
         end
       end
 


aljesusg:
@miq-bot add_label bug

app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb
```diff
@@ -194,7 +194,7 @@ def fetch_server_availabilities(feeds)
           props = server.properties
 
           props['Availability'], props['Calculated Server State'] =
-            process_server_availability(props['Server State'], availability.try(:[], 'data').try(:first))
+            process_server_availability(props['Suspend State'] == 'SUSPENDED' ? 'suspended' : props['Server State'], availability.try(:[], 'data').try(:first))
```
cfcosta: This should be on `process_server_availability`.

aljesusg: The problem is that this function is used by another parser, I though about but in https://github.com/ManageIQ/manageiq-providers-hawkular/blob/c120d9b8306db304c6ea88cf26f5a86bbed0450c/app/models/manageiq/providers/hawkular/middleware_manager/event_catcher/stream.rb#L71 is different so I need to see how to do it.

PR #133, opened 2018-09-12
Add running of common core provider specs

Fryguy:
Add running of common core provider specs


agrare:
The Hawkular provider in master has been removed so this isn't necessary

